PR #4, opened 2015-08-07
Don't always show debug indicators

copperwall:
This adds functionality to only show the debug indicators when the `debug` setting in 
`config.js` is set to `true`.

This feature also allows the user to enter debug mode on the client without having 
to reload the page and lose the current state they have on the page.

To enter debug mode manually, they can open the console and run `App.debugStart()`.
This will set the `App.debug` value to `true` and then rerender the view with the debug indicators.

### DIFF
--- a/controllers/main.js
+++ b/controllers/main.js
@@ -1,12 +1,14 @@
 var socketAuthenticator = require('../lib/socket-auth');
 var socketio = require('../package').dependencies['socket.io'];
+var config = require('../config');
 
 module.exports = {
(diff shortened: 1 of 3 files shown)


andyg0808:
CR :+1:

andyg0808:
QA :+1:

andyg0808:
CR :+1:

andyg0808:
QA :

copperwall:
CR :white_check_mark:

PR #12, opened 2015-09-10
Parse associated Issue in Pull and update from the Issue

andyg0808:
Use the "Closes" and "Connects" keywords to figure out what Issue we should use. Does not write to GitHub, simply updates the Pull object for now.

Issue #7, opened 2015-08-20
Feature Requests


kwiens:
- [ ] On new comments that dev_block, un_dev_block, deploy_block, and un_deploy_block: add or remove `deploy_block` and `dev_block` labels to the pull
- [x] If the pull request doesn't have a milestone and the issue does, add the milestone to the pull request
- [x] Watch for [github 'close' keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) as well as [waffle 'connects'](https://github.com/waffleio/waffle.io/wiki/FAQs#prs-closing-keywords) keyword.
  - Ignore the waffle connects issues for now
- [x] Add Issue, issue title, milestone, pull, difficulty scope, and asignee to new metrics table
- [x] Display the difficulty scope in the UI
- [x] Add filters for assigned and milestone.
  - Ex: assigned=soriat,sterlinghirsh,suzukichi,copperwall,anthonypileggi&milestone=8%2F29

cc @davidrans

kwiens:
Difficulty scope me!

marczych:
> On new comments that dev_block, un_dev_block, deploy_block, and un_deploy_block: add or remove deploy_block and dev_block labels to the pull

Any reason to not do the same for "needs CR", "needs QA", and "ready to deploy"? Having a history of when pulls go into different states would be pretty cool. Also, if I sign off on CR, seeing the "needs CR" label be removed and "ready to deploy" be added I can deploy it immediately rather than opening up pulldasher to see if it's in the ready column.

This definitely doesn't have to be lumped in with the dev_block and deploy_block labels, though.

marczych:
It'd be great to see the CR/QA labels at a glance in Waffle too. We'll get this for free once the labels are applied.

andyg0808:
I just threw this feature in with the dev_block and deploy_block labels. It was three more lines of code.

kwiens:
I think we're good! We did everything that didn't require pulldasher modify Github, and I think that's where we want to leave it.

PR #13, opened 2015-09-11
Add difficulty indicator

andyg0808:
Basic indicator of the current difficulty score on an item, or "?" if there's none set:
image


andyg0808:
deploy_block :+1: on #12

andyg0808:
Connects #7

danielbeardsley:
Rebased onto base pull after merging master and fixing it.

copperwall:
This looks good to me. CR :fire:

danielbeardsley:
Rebased (no conflicts) this to update the commit list and diff.

copperwall:
CR :fire:

controllers/githubHooks.js
```diff
@@ -76,8 +76,11 @@ var HooksController = {
 
          // Update DB with new pull request content.
          dbUpdated = preUpdate.then(function() {
-            return dbManager.updatePull(new Pull(body.pull_request));
-         });
+            var pull = new Pull(body.pull_request);
+            // Update the pull with information from the related issue
+            return pull.syncToIssue();
```
danielbeardsley: I think this logic may be reversed. The name is a bit hard to parse, and perhaps that's why. This function loads the *issue* from the DB and updates the *pull request* with issue data from the DB. To me, it seems like we need to do that when github tells us the issue is updated, not when github tells us the pull is.

lexahall:
dev_block :hand: on merge conflicts

entendu:
I _think_ I resolved the merge conflict properly, please take a look

danielbeardsley:
We typically don't have scores on pulls... we'd need a tool to copy them over or something...

lexahall:
un_dev_block 💯

copperwall:
If story points are now more of an issue thing than a pull request thing, is it worth adding info about it on the page?

colebemis:
I agree with @copperwall. Aren't story points more useful for milestone planning, not CR/QA?

BaseInfinity:
CR 👙

andyg0808:
dev_block :-1: on deciding if we want this.

andyg0808:
It looks like it might not be that hard to support grabbing the size score from the issue when the pull first shows up (at least for pulls that only close one issue). But when the size on an issue is updated, it will likely require a manual refresh on each related pull to show up on them, because the issue doesn't necessarily know which pulls depend on it.

PR #14, opened 2015-09-11
Update GitHub from pulldasher

andyg0808:
When a pull doesn't have a milestone but its associated issue does, add the milestone to the pull.
I've left the code that actually does the thing commented for now, because I've not tested it. I'll uncomment it once I'm fairly confident that it won't break anything.

This also adds a view `pull_list` to the database which makes it easier to see most of the human-useful data we've got on the pulls.

Connect #7
cr_req 2


andyg0808:
deploy_block :+1: on #13

kwiens:
Bump!

marczych:
> When a pull doesn't have a milestone but its associated issue does, add the milestone to the pull.

So pulldasher will update the milestone? It'd be nice if we could leave the permissions as read only (or do we already write data somewhere else too?).

Also, this is a fairly opinionated decision that we might not necessarily want to impose on every pulldasher user.

kwiens:
> Also, this is a fairly opinionated decision that we might not necessarily want to impose on every pulldasher user.

I assume that you're referring to outside organizations — I'm pretty sure that we always want issue and pull milestones synchronized.

I'm OK imposing our will here, but if we wanted to be nice to other folks in the future then we could wrap this in a setting.

danielbeardsley:
> So pulldasher will update the milestone? It'd be nice if we could leave the permissions as read only (or do we already write data somewhere else too?).

Fairly certain we are currently read-only. So, permissions may have to be updated (easy) but this would break the read-only trend.

kwiens:
I understand the concern with permission-creep. I'm going to leave this decision up to @davidrans.

davidrans:
I think I've thought of a happy medium. If the issue which a pull closes/connects is on a milestone, we can display that milestone on the pull blurb without actually adding the pull to the milestone. This also will keep the milestone from getting bloated with duplicates of everything (i.e. `Issue: redesign pulldasher` + `Pull: implement new pulldasher design`).

This way we keep pulldasher read-only (which I really like), but we don't have to throw away all this work @andyg0808 put in and reimplement it somewhere else.

danielbeardsley:
Rebased after fixing up base pull. List of commits here is incorrect (not sure why, list of commits on the CLI looks fine). Diff has the correct stats though.

danielbeardsley:
Rebased on the base pull again (cause it got rebased)

copperwall:
Is this something we even want anymore? I thought that pull request milestones don't really matter now that we only base sprints on issues.

With our current system if a pull request is associated with an issue, and that issue is closed and a new partial is created, do we disassociate that pull with the original issue?

danielj41:
dev_block 👍  on discussion and merge conflicts

kwiens:
> Is this something we want anymore? I thought that pull request milestones don't really matter now that we only base sprints on issues.

Agreed, I don't think we need it. Things are good as-is. Sorry @andyg0808!

Can always re-dust this off if we change our minds.

PR #16, opened 2015-10-17
Milestones: Sort columns based on milestone due date

### LINKED ISSUE #5: Include milestone in sort order
Let's start factoring milestone into the sort order.

Upcoming milestones should be blue, past milestones should be red.

The dev_block date format is 'Jul 19', but we're using the '7/19' format for milestones. Let's make them consistent.

cc @davidrans

copperwall:
This alters the sorting preferences for columns so that pulls with milestones
are placed near the top, and milestones that are past their due date are
sorted even higher towards the top.

Fixes #5

### DIFF
--- a/views/standard/spec/columns.js
+++ b/views/standard/spec/columns.js
@@ -7,7 +7,7 @@ define(['jquery', 'appearanceUtils'], function($, utils) {
 
          if (pull.milestone.title) {
             var due_date = pull.milestone.due_on &&
-            new Date(pull.milestone.due_on);
+             new Date(pull.milestone.due_on);
 
             // If milestone is past due date.
             if (due_date && due_date < new Date()) {
@@ -92,7 +92,7 @@ define(['jquery', 'appearanceUtils'], function($, utils) {
 
                   node.append(link);
                }
-            },
+            }
          },
          shrinkToButton: true
       },
@@ -195,7 +195,7 @@ define(['jquery', 'appearanceUtils'], function($, utils) {
             score -= pull.status.CR.length;
 
             return score;
-         },
+         }
       },
       {
          title: "QA Pulls",
--- a/views/standard/spec/columns.js
+++ b/views/standard/spec/columns.js
@@ -1,4 +1,36 @@
 define(['jquery', 'appearanceUtils'], function($, utils) {
+
+   // Sorting functions that are run on each column.
+   var globalSorts = {
+      sortByMilestone: function(pull) {
+         var score = 0;
+
+         if (pull.milestone.title) {
+            var due_date = pull.milestone.due_on &&
+            new Date(pull.milestone.due_on);
+
+            // If milestone is past due date.
+            if (due_date && due_date < new Date()) {
+               score -= 1500;
+            } else {
+               score -= 750;
+            }
+         }
+
+         return score;
+      }
+   };
+
+   /**
+    * Runs all of the functions from globalSorts on the given pull and reduces
+    * each score into a total sorting score.
+    */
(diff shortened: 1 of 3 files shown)


views/standard/spec/columns.js
```diff
@@ -179,6 +203,18 @@ define(['jquery', 'appearanceUtils'], function($, utils) {
...
+                new Date(pull.milestone.due_on);
+
+               // If milestone is past due date.
+               if (due_date && due_date < new Date()) {
+                  score -= 2000;
+               } else {
+                  score -= 1000;
+               }
+            }
+
```
marczych: Is there an easy way to reuse this logic rather than duplicating it a few times?

copperwall: @marczych there may be a way to add the repeated logic to a generic column sort method. I'll see if it's doable. 

dev_block :star:

copperwall: Aha! As long as this function does what I think it does, I should be able to add the logic to one place.

https://github.com/iFixit/pulldasher/blob/master/views/standard/spec/index.js#L33

copperwall:
@marczych I put the the milestone sorting logic in a `sortByMilestone` function in a `globalSorts` object in `spec/columns`.

Each column gets its initial sorting score by running the given pull through each column function and reducing each of their scores into one global sort score.

This way if we have any other common sorting methods across columns we can add an extra function to `globalSorts` and it will be applied to all columns from a single place.

un_dev_block :star:

copperwall:
I tested this here
image

So I think we're good for QA

solflare:
CR :moon: Will this cause our own pulls to be pushed down by pulls that are past due?

copperwall:
@solflare It will for the dev blocked column.

The priority will be

1. Past due pulls that belong to the user
1. Past due pulls
1. Pulls on a milestone that belong to the user
1. Pulls on a milestone
1. Pulls that belong to the user that aren't on a milestone
1. The rest should be sorted by the amount of time they've been dev blocked

The CR and QA columns already sort pulls that belong to the current user towards the bottom of the list.

solflare:
Ahh that makes sense, nice job!

copperwall:
Thank you!

views/standard/spec/columns.js
```diff
@@ -1,4 +1,36 @@
...
+         if (pull.milestone.title) {
+            var due_date = pull.milestone.due_on &&
+             new Date(pull.milestone.due_on);
+
+            // If milestone is past due date.
+            if (due_date && due_date < new Date()) {
+               score -= 1500;
+            } else {
+               score -= 750;
```
BaseInfinity: Where are these score coming from?

copperwall: They are just arbitrary numbers that are relative to other arbitrary numbers currently used in the other sorting functions. It would probably be a good idea to set a fixed scale of point values to choose from instead of just throwing some in that work.

I chose these in particular because it would override any other of the scores from the other sorts.

I have issue #17 open for suggestions for improving the sort feature. It would define a list of sorting functions (i.e. `is_mine`, `needs_cr`, `needs_qa`) that would be listed in the column's `sort` method. Then the sum of all the scores could be the pull's score for that column. That way there is a centralized number of sorting methods that a column can use.

Maybe we could include a multiplier for the score, like how CR and QA columns both factor in whether or not there is QA and CR needed, but the CR column could have a 5x multiplier on the `needs_cr` sort while the QA column could have a 5x multiplier on the `needs_qa` method.

sterlinghirsh:
CR :zap:

PR #27, opened 2016-04-28
Example config: update to handle unicode emojis as well

### LINKED ISSUE #26: signatures: accept unicode emoji
[`signatures--accept-unicode-emoji`](https://github.com/iFixit/pulldasher/compare/signatures--accept-unicode-emoji)

danielbeardsley:
Github started making their emoji auto-completor inject unicode emojis
instead of ascii representations like :+1: or :smile:. This updates
the regex to detect unicode ranges that contain emojis. The regex is a
bit more permissive than needed, but the set symbols that are used
as emojis is dispersed across a smattering of unicode ranges.

Closes #26

### DIFF
--- a/config.example.js
+++ b/config.example.js
@@ -1,4 +1,10 @@
 var port = process.env.PORT || 3000;
+
+// Support text emojis and almost all unicode emojis / symbols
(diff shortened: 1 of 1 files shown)


copperwall:
This looks fine to me. CR :bowtie:

PR #31, opened 2016-05-14
Pull Request Diff Comments: pay attention!

### LINKED ISSUE #28: Comments: pay attention to pull request review comments
[`pull-request-review-comments`](https://github.com/iFixit/pulldasher/compare/pull-request-review-comments)

We currently listen to and record only *issue* comments (as they are called). These are the comments at the bottom of a pull request, not associated with a particular diff. It would be pretty trivial to also listen to [pull request review comments](https://developer.github.com/v3/activity/events/types/#pullrequestreviewcommentevent) (the comments on a diff).  There are often more review comments than issue comments, so it's an important source of data.

The minimal data we store about issue comments (id, number, author, ...) is all available on review comments, so we could even use the same table (we'd have to ad a type column to the PK though).

Wether we parse these comments for signatures is a separate discussion.

- Add type column to comments (make it part of PK)
- Listen to new event type and create Comment records.

danielbeardsley:
Now, all comments (pull and pull-review) are stored in the DB
and on the Pull objects in the Pull.comments array.

This will be useful for analysis on the DB side and could be useful on
the client side to remind you which pulls you've been active on.

Note: I've added our first migration. Dealing with them should be
pretty simple, just run (manually) the ones that haven't been run.

Closes #28

### DIFF
--- a/lib/git-manager.js
+++ b/lib/git-manager.js
@@ -72,6 +72,7 @@ module.exports = {
       // before errors happen.
       githubPull.body = githubPull.body || '';
 
+      var reviewComments = module.exports.getPullReviewComments(githubPull.number);
       var comments = module.exports.getPullComments(githubPull.number);
       var headCommit = module.exports.getCommit(githubPull.head.sha);
       var commitStatus = module.exports.getCommitStatus(githubPull.head.sha);
@@ -79,12 +80,13 @@ module.exports = {
 
       // Returned to the map function. Each element of githubPulls maps to
       // a promise that resolves to a Pull.
-      return Promise.all([comments, headCommit, commitStatus, events])
+      return Promise.all([reviewComments, comments, headCommit, commitStatus, events])
        .then(function(results) {
-         var comments = results[0],
-             headCommit = results[1],
-             commitStatus = results[2],
-             events = results[3];
+         var reviewComments = results[0],
+             comments       = results[1],
+             headCommit     = results[2],
+             commitStatus   = results[3],
+             events         = results[4];
 
          // Array of Signature objects.
          var signatures = comments.reduce(function(sigs, comment) {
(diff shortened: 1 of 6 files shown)


danielbeardsley:
FYI, this is QA 👍 

I was running the server as I was developing.

If anyone wants to see this running, talk to me.

sctice:
Great! CR 🎏

lib/git-manager.js
```diff
@@ -72,19 +72,21 @@ module.exports = {
...
        .then(function(results) {
-         var comments = results[0],
-             headCommit = results[1],
-             commitStatus = results[2],
-             events = results[3];
+         var reviewComments = results[0],
+             comments       = results[1],
+             headCommit     = results[2],
+             commitStatus   = results[3],
+             events         = results[4];
```
BaseInfinity: Will changing the order of these break anything that relied on this order?

lib/git-manager.js
```diff
@@ -147,6 +159,17 @@ module.exports = {
...
       }).then(getAllPages);
    },
+
+   getPullReviewComments: function(pullNumber) {
+      // "review" comments are the comments on the diff of pull
+      var getPullComments = Promise.denodeify(github.pullRequests.getComments);
+      return getPullComments({
+         user: config.repo.owner,
+         repo: config.repo.name,
+         number: pullNumber
```
BaseInfinity: This could prob be DRYd up (noticed we copy pasta the config in each of these functions) but for a future pull

BaseInfinity:
CR :bikini:

lib/git-manager.js
copperwall: I think this will be fine, since the results array was only used for binding promise results to local variables. The return value from `parse` is a new `Pull` object, so nothing should be using `results` after this function returns.

copperwall:
CR :three:

PR #32, opened 2016-05-14
Pulldasher UI Update

soriat:
Not much functionality has changed, but the UI has been freshened up and a lot of code has been cleaned up.

Pictures
==
 * Tony's Mockup: http://i.imgur.com/iVfm5YE.png
 * New Day Theme vs Old Theme: http://i.imgur.com/PlXffAS.png
 * Day Theme vs Night Theme: http://i.imgur.com/7MvOmSd.png

Main Changes
==
 * Theme switcher! The only new theme so far is the Dark Theme, but adding new ones is trivial now.
 * Pulls are sorted more intelligently by default.
  * Your pulls will always be at the top of each column followed by milestones (oldest first)
 * Indicators are now icons instead of labels.

QA
==
Since pulldasher is pointing at this branch instead of master, I suppose everyone is QAing this!


views/standard/index.html
```diff
@@ -5,14 +5,57 @@
...
+
+      // Color switcher
+      $(document).ready(function() {
+         var cscookie = "";
+         document.cookie.split(';').forEach(function(currentValue) {
+            var elem = currentValue.split('=');
+            if (elem[0] == "pulldasher-colorscheme") {
+               cscookie = elem[1];
+            }
+         });
```
danielbeardsley: All these loaded libraries and none of them has a cookie parser?

danielbeardsley:
This is pretty!

But I have some usability complaints:

* The indicator for "My Pull" used to be a lot more visible. I appreciate being able to see where my pulls are in the line-up and putting the icon on the left side with all the busyness over there makes it less useful.
image

* The build status indicators, while fancy and animated, are less useful because the important ones (failed states) don't stand out as much. We previously hid the passing state until you `:hover`  so the non-passing ones would stand out. Perhaps add one more failed state (No build status at all) then hide the passing state until hover.

* The "Refresh" button is super prominent but is almost never used / needed.
image

* I think seeing the milestone date all the time was useful.

danielbeardsley:
Also, you hadn't merged master in since a few pulls back so deploying this branch meant losing a few recent pulls. I merged and re-deployed.

controllers/githubHooks.js
```diff
@@ -91,7 +92,7 @@ var HooksController = {
          body.comment.number = body.pull_request.number;
          body.comment.repo =   body.repository.name;
          body.comment.type =   'review';
-         var comment = new Comment(body.comment);
```
copperwall: Thanks hoisting.

views/standard/index.html
copperwall: jQuery has a third party cookie plugin, but that's more library code to load :X

soriat: I'm sure there are plenty, but @suzukichi wrote this and it works great.

suzukichi: During the hackathon, I couldn't find if any of our libraries did it automatically. I wasn't feeling so hot though, so I just put this together and went home. 

I'm not morally opposed to anyone fixing it if they find a cookie parser in one of the libraries. 
Jquery's is 3rd party though, like copperwall said, and we don't include that anywhere atm.

soriat:
@danielbeardsley 
> The indicator for "My Pull" used to be a lot more visible

I'm fairly certain it was basically the same icon, but a slightly different color. Since all your pulls are guaranteed to be at the top of every column anyway, what more do you want?

> We previously hid the passing state... (and that was better)

You're right, that's fixed now.

> The "Refresh" button is super prominent but is almost never used / needed.

It's a tiny circle that's hidden until you hover over the pull... so I wouldn't really call it prominent. As for it "never being used / needed" I think that depends on the user. If we need the real estate for something else or nobody at iFixit actually uses it then yeah, we should remove it.

> you hadn't merged master

Oops, I wasn't aware the branch was out of date. That said... I was baffled by how many typos were introduced: 7bcb1ce. How did that happen?

danielbeardsley:
> I'm fairly certain it was basically the same icon, but a slightly different color. Since all your pulls are guaranteed to be at the top of every column anyway, what more do you want?

It was the same or similar, but it used to stand out a lot more / maybe used to be on the right side? Also, no, they don't show up at the top, they are seemingly scattered or at the bottom. Perhaps the indicator could be right-aligned or something. Perhaps a better solution is to implement more toggles that use css classes to hide / show various pulls.

Typos? I didn't see any typo fixes in that commit. Just adding a bunch of semi-colons.

PR #34, opened 2016-05-18
Github: obey the rate limit!

### LINKED ISSUE #33: github api: obey speed limit
[`github-api--obey-speed-limit`](https://github.com/iFixit/pulldasher/compare/github-api--obey-speed-limit)

danielbeardsley:
All hail the rate limit!

This adds some smarts to our github usage and will allow requests to
go through unimpeded until we get to the choke point, at which point
we start throttling requests such that we stay *just* under the rate
limit.

Obviously, not knowing the future means we can't be all that smart
about it.

The delay is such that if you have an unlimited number of requests
coming in, we'll stay just below the rate limit. If you reach the
choke point but only have a few more requests, they will be throttled
unnecessarily, but we'll just have to live with that.

This means that now, the rebuild-history option will totally work
without hitting the rate limit. Also, change rebuild-history to
operate on pulls one at a time, instead of thousands in parallel.

Closes #33

### DIFF
--- a/app.js
+++ b/app.js
@@ -14,6 +14,7 @@ var config = require('./config'),
     Signature = require('./models/signature'),
     mainController = require('./controllers/main'),
     hooksController = require('./controllers/githubHooks'),
+    debug = require('debug')('pulldasher'),
     reqLogger = require('debug')('pulldasher:server:request');
 
 var args = process.argv.slice(2);
@@ -120,27 +121,33 @@ function refreshAll(all) {
    var githubPulls;
 
    if (all === true) {
+      debug("rebuilding all pulls");
       githubPulls = gitManager.getAllPulls();
    } else {
+      debug("rebuilding all open pulls");
       githubPulls = gitManager.getOpenPulls();
    }
 
-   queue.pause();
-
    githubPulls.done(function(gPulls) {
-      var pullsUpdated = gPulls.map(function(gPull) {
+      debug("done retrieving pulls");
+      function next() {
+         if (!gPulls.length) {
+            debug("done building all pulls");
+            return;
+         }
+
+         var gPull = gPulls.shift();
+         debug("building pull %s", gPull.number);
+
+         queue.pause();
          var pull = gitManager.parse(gPull);
 
-         return new Promise(function(resolve, reject) {
-            update(pull).done(function() {
-               resolve();
-            });
+         update(pull).done(function() {
+            queue.resume();
+            next();
          });
-      });
-
-      Promise.all(pullsUpdated).done(function() {
-         queue.resume();
-      });
+      };
+      next();
    });
 }
 
--- a/lib/git-manager.js
+++ b/lib/git-manager.js
@@ -6,20 +6,32 @@ var GithubApi = require('github'),
       debug: config.debug,
       version: '3.0.0'
     }),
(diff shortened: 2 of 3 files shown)


app.js
```diff
@@ -120,27 +121,33 @@ function refreshAll(all) {
    var githubPulls;
 
    if (all === true) {
+      debug("rebuilding all pulls");
```
BaseInfinity: Not sure if we want to keep these debug statements around but it is kinda nice keeping it around so I'm indifferent about it

lib/git-manager.js
```diff
@@ -6,20 +6,32 @@ var GithubApi = require('github'),
...
+    rateLimit = require('./rate-limit.js');
 
 github.authenticate({
    type: 'oauth',
    token: config.github.token
 });
 
-var get    = Promise.denodeify(github.pullRequests.get);
-var getAll = Promise.denodeify(github.pullRequests.getAll);
+function denodeify(func) {
```
BaseInfinity: I wonder if we should change this function name to show that we are running these functions through rateLimit object

lib/git-manager.js
```diff
@@ -161,17 +168,14 @@ module.exports = {
    },
 
    getPullReviewComments: function(pullNumber) {
-      // "review" comments are the comments on the diff of pull
-      var getPullComments = Promise.denodeify(github.pullRequests.getComments);
```
BaseInfinity: me likey

lib/rate-limit.js
```diff
@@ -0,0 +1,85 @@
+var Promise = require('promise'),
+    debug = require('debug')('pulldasher:rate-limit');
```
BaseInfinity: there might be a module for this but this is short and concise so im cool with this

BaseInfinity:
left small comments, but otherwise CR 👙 this is schweet

app.js
danielbeardsley: The debug module is what we use for debug output (throughout the whole project).

`DEBUG='*' bin/pulldasher` to see all output.

lib/git-manager.js
danielbeardsley: Thought about it... but it's too wordy and the function is defined right there. `rateLimitedDenodify` ?

lib/rate-limit.js
danielbeardsley: I don't get it. `debug` *is* a module. We're using it...

danielbeardsley:
Also this is QA 👍  I've run it a bunch while working on it and examined the DB.

solflare:
CR :moon:

lib/git-manager.js
BaseInfinity: Lol ya I feel the same way

lib/rate-limit.js
BaseInfinity: Sorry, bad place to leave a comment. I meant a rate-limiting module

danielbeardsley: The only *rate limiting* part of the code that isn't specific to our use case is the last 25 lines of this file (the rest deals with github headers, choke points, ...).

Never the less, every one I looked at either through away extra calls (i.e. make 10 calls == only 2 get through) and / or had a fixed time interval on creation. We want our interval to be flexible and respond to the amount of time / requests remaining.

PR #35, opened 2016-05-20
Github Api: bunch of rate-limit fixes

danielbeardsley:
There some unpolished parts of the recent rate limit pull, as well as
some bugs we missed during CR.

The individual commits offer a more detailed history.

Summary:

* Api: calculate rate limit safely, prevent -1 timeout
* Pull Manager: forget about closed pulls
* Api: use 100 as default page size
* Api: DRY up default params
* Api: make getAllPages get ALL, not just the first 5
* Api Rate limit: don't rate-limit till there are only 200 left
* Migrations: move into subdirectory, forgot to do that


lib/rate-limit.js
```diff
@@ -81,5 +81,9 @@ function startDrainer() {
 }
 
 function interval() {
-   return (resetAt * 1000 - Date.now()) / (remaining + 1);;
+   var requestsRemaining = Math.max(1, remaining);
+   // Always underconsume, and leave a minute's worth of requests remaining
+   // just in case our clock drifts from githubs
+   var timeTillReset = Math.max(1, (60 + resetAt) * 1000 - Date.now());
```
davidrans: I don't think `resetAt` is defined.

davidrans:
CR :coffee: minus https://github.com/iFixit/pulldasher/pull/35#discussion_r65100038

lib/rate-limit.js
danielbeardsley: Yeah it is. https://github.com/iFixit/pulldasher/blob/83410a041d0dc27f0749da43280b3f98481598d4/lib/rate-limit.js#L28

davidrans: Gotcha, I was confused by the global variables. Is [`resolve`](https://github.com/iFixit/pulldasher/pull/35/files#diff-80d900d5fc6a583850c7b52862522c34R62) supposed to be a global variable? I don't see where that's defined.

danielbeardsley: Fixed.

davidrans:
CR :coffee:

davidrans:
CR :coffee:

PR #11, opened 2015-09-10
Add issues support to Pulldasher

andyg0808:
Makes Pulldasher support receiving information about issues and writing
that information to the database. This allows analysis of issues with
SQL.

Connect #7

PR #37, opened 2016-05-31
Github Comment events: deal with edit / delete

### LINKED ISSUE #36: Comment editing produces duplicate signatures
Editing a comment with a signature duplicates that signature in the UI and the database.

Bad.

image

Github now sends *edit* events when comments are edited, when processing signatures, we should be doing `INSERT OR REPLACE` or something.

danielbeardsley:
Now we can finally deal with these new event types.
(after all the preparatory work in this branch)

Since pull request review comments can't have signatures,
we only need to delete or update the comment.

On the other hand, issue-level comments on pulls can have signatures
and if we edit a comment to introduce or remove a signature, that
could have large affects on the rest of the signatures and there's no
easy way to undo the changes that an un_dev_block signature has done.
The easiest way to deal with it is to completely refresh the given
pull. It's a little overkill, but it's very little code and it's easy
to understand.

Closes #36


danielbeardsley:
deploy_block 👍  on pull #11 being merged.

copperwall:
CR :fire:

PR #44, opened 2016-07-10
Pulldasher Install: Fix issues causing jshint to fail.

solflare:
These things were causing pulldasher to fail on install when starting up a new dev machine. For the most part the changes are straightforward but I wasn't quite sure what we wanted to do about the unused variables so for now I removed those checks from jshint but I can change it to just removing the offending variables.

CC @danielbeardsley


solflare:
Not sure if we need qa but QA :ididthething:

soriat:
CR ⚡

davidrans:
I think we should remove the unused variables, not the jshint check for them. Unused variables are never intentional, and sometimes cause bugs (e.g. if it's a typo'd variable).

dev_block :hand:

solflare:
Sure thing, I wasn't sure if some of them were things Danny was keeping around but they can definitely be re-added when needed.

solflare:
un_dev_block :moon:

danielbeardsley:
Thanks @solflare !!

danielbeardsley:
CR 👍

davidrans:
CR ☕

PR #46, opened 2016-08-29
Cleanup: fix some recently introduced small bugs

### LINKED ISSUE #45: issues syncing: cleanup
[`issues-syncing--cleanup`](https://github.com/iFixit/pulldasher/compare/issues-syncing--cleanup)

danielbeardsley:
A couple of pulls were merged recently with small (though breaking)
issues because they didn't get enough code review (or testing)

This addresses the known bugs/regressions.

* Disable pull milestone syncing
   * Good idea, but it pretty much isn't finished
   * There is a pull that gets us closer
* Re-add a few missing variables
   * They were removed after a js-hint cleanup, but there was an open
     pull which added some usages.
* Re-add a missing function
   * Still not sure why it was removed without comment.

Closes #45


models/pull.js
```diff
@@ -75,6 +75,9 @@ Pull.prototype.update = function() {
 };
 
 Pull.prototype.syncToIssue = function() {
+   return Promise.resolve(this);
```
solflare: Just double checking does this need to be 
```js
return new Promise.resolve(this);
```

davidrans:
CR ☕

models/pull.js
danielbeardsley: Nope, [`Promise.resolve()`](https://www.npmjs.com/package/promise#promiseresolvevalue) is a static function, not a thing you should instantiate.

solflare:
CR :moon:

PR #54, opened 2016-10-23
Add Dockerfile

copperwall:
This adds a Dockerfile that allows running pulldasher in a docker
container. The container does not include an instance of mysqld, so the
pulldasher container will not be able to use a connection to mysql using
localhost. It will have to be able to connect to a mysqld server running
in either another container, or another ip address.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM node:latest
+
+RUN mkdir -p /opt/pulldasher
+WORKDIR /opt/pulldasher
+
+# Install dependencies
+COPY package.json /opt/pulldasher
+RUN npm install
+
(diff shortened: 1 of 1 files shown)


kwiens:
Cool! Is this an experiment, or do you plan to start using it?

copperwall:
Mostly an experiment! Also I thought it'd be nice to have a default way for people to safely develop this on their own computers. It's a little more work for people on OSX, but it looks like they've made it easier to use Docker on OSX using xhyve instead of virtualbox.

Maybe we could use pulldasher as a trial run for versioning and storing the project as different docker images. I think this is also good to go for actually running production pulldasher, so I can try transitioning to that when I have a little more free time.

PR #56, opened 2016-11-01
/fonts path: serve the correct directory

### LINKED ISSUE #55: fonts: use correct path
[`fonts--use-correct-path`](https://github.com/iFixit/pulldasher/compare/fonts--use-correct-path)

danielbeardsley:
The only fonts we use are the font-awesome ones where are not present
in the bootstrap fonts dir. It seems we dont' use the bootstrap glyph
fonts at all so it's Ok to change this dir.

Before, we were getting a 404 when looking for font-awesome files in
the /fonts dir. After, they exist!

Closes #55


kwiens:
Should we merge this?

danielbeardsley:
Yes.

PR #61, opened 2017-01-10
Github API Responses: don't depend on user.login

### LINKED ISSUE #59: Bug: crash in git-manager
[`event-actors--handle-nulls`](https://github.com/iFixit/pulldasher/compare/event-actors--handle-nulls)

Pulldasher crashed, here's the relevant log entry:
```
 /opt/pulldasher/node_modules/promise/node_modules/asap/asap.js:45
                 throw e;
                 ^
 TypeError: Cannot read property 'login' of null
     at /opt/pulldasher/lib/git-manager.js:262:27
     at Array.map (native)
     at getLabelsFromEvents (/opt/pulldasher/lib/git-manager.js:258:20)
     at /opt/pulldasher/lib/git-manager.js:196:23
     at /opt/pulldasher/node_modules/promise/core.js:33:15
     at flush (/opt/pulldasher/node_modules/promise/node_modules/asap/asap.js:27:13)
     at nextTickCallbackWith0Args (node.js:420:9)
     at process._tickCallback (node.js:349:13)
```

danielbeardsley:
Sometimes github user accounts are deleted.
When they are included in an api response, the user looks like:

Normal User:

    issue: {
        user : {
            login: "danielbeardsley",
            url: ...
        }
    }

Deleted User:

    issue: {
        user : null
    }

This change wraps all places where we access properties of the user
object in an api response.

Most places would have previously crashed the app. The one that
changed now produces "ghost" instead of NULL (just like github's
web UI). This allows the DB columns to still be NOT NULL. Keeping them
from being null makes it so we don't have to deal with nulls in the
rest of our code.

Perhaps in the future we'll want to do something different, but
this is safe for now.

Note: I found these by:
git grep -E "\.(sender|actor|user|assignee)"

And looking at the backend files that matched. The frontend files
dont' have to change because they only deal with our model objects
that have been constructed from API responses (which now inject this
ghost user).

Closes #59

### DIFF
--- a/models/db_signature.js
+++ b/models/db_signature.js
@@ -1,4 +1,5 @@
 var utils = require('../lib/utils'),
+    getLogin = require('../lib/get-user-login'),
     db = require('../lib/db');
 
 /**
@@ -9,7 +10,7 @@ function DBSignature(signature) {
    var sigData = signature.data;
    this.data = {
       number:     sigData.number,
-      user:       sigData.user.login,
+      user:       getLogin(sigData.user),
       userid:     sigData.user.id,
       type:       sigData.type,
       date:       utils.toUnixTime(sigData.created_at),
--- a/models/db_signature.js
+++ b/models/db_signature.js
@@ -1,5 +1,6 @@
 var utils = require('../lib/utils'),
     getLogin = require('../lib/get-user-login'),
+    getUserid = require('../lib/get-user-id'),
     db = require('../lib/db');
 
 /**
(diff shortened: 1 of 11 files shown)


cdcline:
CR 👍 
@danielbeardsley merge conflicts?

models/db_signature.js
```diff
@@ -1,4 +1,6 @@
 var utils = require('../lib/utils'),
+    getLogin = require('../lib/get-user-login'),
+    getUserid = require('../lib/get-user-id'),
```
cdcline: I don't see `get-user-id`?
image

danielbeardsley: Correct. Just pushed the fix. Sorry!

cdcline:
CR 👍

copperwall:
LGTM CR :fire:

PR #63, opened 2017-01-31
Startup: Load open pulls from the DB

### LINKED ISSUE #60: Pulls: load pulls from DB on startup.
We used to load all the open pulls from the DB upon startup and then refresh them in the background. This prevents pulldasher from starting up blank when it restarts.

It's a simple change and probably just got dropped in all the refresh refactoring.

danielbeardsley:
So we don't start out blank.

There's a possibility that we'll load pulls that have been
closed in the meantime, but it's probably better that we
know about them anyway.

Closes #60

### DIFF
--- a/app.js
+++ b/app.js
@@ -7,6 +7,8 @@ var config = require('./config'),
     socketAuthenticator = require('./lib/socket-auth'),
     refresh = require('./lib/refresh.js'),
     pullManager = require('./lib/pull-manager'),
+    dbManager = require('./lib/db-manager'),
+    pullQueue  = require('./lib/pull-queue'),
     mainController = require('./controllers/main'),
     hooksController = require('./controllers/githubHooks'),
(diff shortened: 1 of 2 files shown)


copperwall:
This looks good to me. CR :fire:

danielbeardsley:
Made more commits after actually testing this. Might want to re-CR since I can't be trusted :-)

bin/pulldasher
```diff
@@ -5,7 +5,6 @@ var debug = require('debug')('pulldasher:bin');
 
 // Save the pid to a file if requested (for an init script)
 if (config.pidFile) {
-   fs.writeFileSync(config.pidFile, process.pid);
```
copperwall: Why are we getting rid of the pidFile functionality?

danielbeardsley: Cause I'm dumb! Fixing.

copperwall:
Cool! CR :fire:

PR #66, opened 2017-02-11
Add docker-compose configuration

copperwall:
This adds a `docker-compose.yml`, which specifies
how the pulldasher_web and pulldasher_db containers will need to
interact with each other.

With docker-compose, you can "one-click" start a Pulldasher instance on
your local machine with a local database.

Here's a quick specification of how it works:

`web` and `db` are two separate containers. The `web` container uses the
image built by running the Dockerfile that comes with Pulldasher. This
container starts by running `entrypoint.sh`, which makes sure the
database is configured with the correct table before starting
pulldasher. It also mounts the current directory within the container so
that any changes made on the host repo are reflecting in the container.
Port 8080 on the container is mapped to the host's 8080, so that you can
access pulldasher at localhost:8080.

`db` specifies that we are using the latest MariaDB image, specifies
that we are mounting a persistent volume `db_data` to `/var/lib/mysql`
and specifies which environment variables to start MariaDB with.


solflare:
CR :moon:

djmetzle:
Not too shabby CR :bangbang:

PR #67, opened 2017-02-17
Update npm dependency versions

copperwall:
This bumps the versions of middleware and standalone npm packages that we use
throughout pulldasher.

This only covers low-hanging fruit modules that don't have breaking changes.
Once example of a module that isn't getting updated in this pull is Express,
because moving from v3 to v4 is going to require some code changes.

This also fixes a possible bug found by JSHint where we were using a variable
before it was actually assigned.


solflare:
CR :moon:

PR #70, opened 2017-02-19
Update app.js to use Express v4

### LINKED ISSUE #69: Upgrade to Express 4
Express 3 is no longer the current major version, so we should try to upgrade it.

The main breaking change is that Express 4 doesn't include the same middleware that Express 3 does. The fix for this is including equivalent middlewares and using those instead.

New packages needed:
* body-parser https://www.npmjs.com/package/body-parser
* cookie-parser
* express-session

copperwall:
This includes the new middleware packages that are no longer included in
Express v4. cookie-parser and cookie-session are no longer necessary due to
express-session.

Closes #69


solflare:
CR :moon:

PR #73, opened 2017-03-01
Redirect to original URL after OAuth success

copperwall:
Previously if you accessed pulldasher using a query param like `assigned` or
`milestone` and you were placed in an OAuth redirect, your query params would
be forgetten after returning from OAuth redirects.

This stores the original URL in the session so that when the user gets back
from an OAuth redirect, the app redirects them to their original URL and not
just `'/'`.

CC @sctice

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -6,6 +6,7 @@ module.exports = function(grunt) {
     // Task configuration.
     jshint: {
       options: {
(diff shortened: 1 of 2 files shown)


kimarre:
CR 💐

sctice:
Sweet! Thanks 😁

PR #78, opened 2017-05-14
Get it working: Correct dependencies

danielbeardsley:
Use the correct dependencies and drop the only `let` usage.

Seem commits for more info.

### DIFF
--- a/lib/authentication.js
+++ b/lib/authentication.js
@@ -48,9 +48,10 @@ module.exports = {
                   debug('Got login from ' + user.username);
 
                   // Redirect user to original URL. Fallback to '/' if DNE.
+                  var redirectUrl = req.session.auth_redirect;
                   delete req.session.auth_redirect;
 
-                  res.redirect(req.session.auth_redirect || '/');
+                  res.redirect(redirectUrl || '/');
                });
             });
          })(req,res,next);
--- a/lib/authentication.js
+++ b/lib/authentication.js
@@ -37,12 +37,12 @@ module.exports = {
          function(err, user, info) {
             if (err) {
                debug("Error: " + err);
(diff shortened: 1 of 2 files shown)


lib/authentication.js
```diff
@@ -48,10 +48,9 @@ module.exports = {
                   debug('Got login from ' + user.username);
 
                   // Redirect user to original URL. Fallback to '/' if DNE.
-                  let redirectUrl = req.session.auth_redirect;
                   delete req.session.auth_redirect;
 
-                  res.redirect(redirectUrl || '/');
+                  res.redirect(req.session.auth_redirect || '/');
```
copperwall: This uses `req.session.auth_redirect` after the field is deleted D:

danielbeardsley: I'm a dum-dum. Thanks!

danielbeardsley:
CR 👍

lib/authentication.js
```diff
@@ -40,6 +40,11 @@ module.exports = {
                next(err);
             }
 
+            if (user === undefined) {
+               debug("Bad user");
+               return;
```
danielbeardsley: I think we need to also either respond or call `next()`

lib/authentication.js
```diff
@@ -40,6 +40,11 @@ module.exports = {
                next(err);
```
danielbeardsley: Hmm.. shouldn't this be `return next()`? Maybe that's the source of the problem?

copperwall: Hmm maybe? I'll check it out.

danielbeardsley:
CR 👍

danielbeardsley:
QA 👍

PR #85, opened 2017-08-03
Search GitHub org for repos

andyg0808:
Rather than requiring each repo to be listed, automatically search for
all repos in a GitHub organization and watch them.

QA
---
These changes are currently completely untested.

### DIFF
--- a/app.js
+++ b/app.js
@@ -9,6 +9,7 @@ var config = require('./config'),
     refresh = require('./lib/refresh'),
     pullManager = require('./lib/pull-manager'),
     dbManager = require('./lib/db-manager'),
+    gitManager = require('./git-manager'),
     pullQueue  = require('./lib/pull-queue'),
     mainController = require('./controllers/main'),
     hooksController = require('./controllers/githubHooks'),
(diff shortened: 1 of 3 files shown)


BaseInfinity:
CR 👙

danielbeardsley:
Maybe make this optional? I'm not sure we want issues from *every* repo we have. Also, this doesn't watch them (you have to add hooks for that), it merely refreshes open pulls on restart.

saltaverde:
CR 🕺

andyg0808:
dev_block :+1: on Danny's comment.

andyg0808:
@danielbeardsley How would you feel about a blacklist?

As far as hooks, how do we feel about add polling support for repos we don't have hooks for? I think having Pulldasher update from them once a day would be better than nothing, as it would help to ensure pulls on those repos didn't get lost and forgotten.

CC: @entendu

danielbeardsley:
> how do we feel about add polling support for repos we don't have hooks for?

There not any reasonable way to do that... once a day query *all* issues and *all* pull requests?

Better just to use hooks and stay up to date. If anything, I'd rather have a script that injects the correct hooks (via the api) for a given repo so it's easy to add them.

> How would you feel about a blacklist?

Not great. If we're a use-case, we'd probably have 2-3 times as many repos on the blacklist as we'd want included. We have many public forks as a backup where we never intend on using pulls or issues.

entendu:
I prefer a whitelist

andyg0808:
Since this won't work in its current condition, and there's no clean way to make it work, I move we close it. All in favor?

danielbeardsley:
Favor!

PR #83, opened 2017-06-30
Remove repo name

MooreGuy:
Now that repo is the new column that is being used in all of the tables, the old repo_name and repo_owner columns can be dropped for good.

QA: Run the migrations, then test all the normal features of pulldasher, signing off and having it update, moving pulls between columns, and test all the links to pulls and their build statuses

Closes: https://github.com/iFixit/pulldasher/issues/79

PR #92, opened 2017-08-09
Drop repo_name column and update schema

MooreGuy:
repo_name is being replaced by repo since we want both the owner and
repo name. Code changes to stop referencing the repo_name happen in
https://github.com/iFixit/pulldasher/pull/83

### DIFF
--- a/migrations/0009-allow-null-repo-names.sql
+++ b/migrations/0009-allow-null-repo-names.sql
@@ -1,6 +1,2 @@
-ALTER TABLE comments MODIFY COLUMN repo_name varchar(255) NULL;
-ALTER TABLE pull_labels MODIFY COLUMN repo_name varchar(255);
-ALTER TABLE pulls MODIFY COLUMN repo_name varchar(255);
-
 ALTER TABLE pulls DROP KEY pulls_repo;
 ALTER TABLE pulls ADD KEY pulls_repo (repo);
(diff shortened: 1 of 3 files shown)


copperwall:
What about the schema :(

lexahall:
QA ?

MooreGuy:
@copperwall Yes! Thanks, I'll get that, also yeah, I'll have to think of QA for this, it touches a lot. dev_block :+1:

MooreGuy:
un_dev_block :+1:

migrations/schema.sql
```diff
@@ -140,7 +137,7 @@ CREATE TABLE IF NOT EXISTS `pulls` (
   `difficulty` int(11) DEFAULT NULL,
   PRIMARY KEY (`repo`,`number`),
   KEY `pulls_state` (`state`),
-  KEY `pulls_repo` (`repo_owner`,`repo_name`),
+  KEY `pulls_repo` (`repo_owner`),
```
nvaa36: It looks like the migration changes from the first one were not added. This is supposed to be a key on column `repo`, not `repo_owner`, right?

lexahall:
@MooreGuy do you have any ideas for QA yet?

cdcline:
CR 👍

copperwall:
dev_block :star:  on https://github.com/iFixit/pulldasher/pull/92#discussion_r137600237

MooreGuy:
I got slightly frustrated with splitting this into two pulls so this pull has all of these changes plus more: https://github.com/iFixit/pulldasher/pull/109

PR #95, opened 2017-08-18
Docker Compose: Make DB initialization idempotent

copperwall:
There's an issue where you can't run `docker-compose` on an already
initialized db, because the `migrate` script tries to create tables that
already exist. By modifying the `dump-schema` command to use `CREATE
TABLE IF NOT EXIST` statements, bringing up pulldasher instance with and
already initialized db with `docker-compose` will work without failure.

This also moves some one-off scripts into the `bin` directory so that
the app root directory looks a little less cluttered.

### QA

Bring up a pulldasher instance using `docker-compose`, make sure it starts without issue.
Kill the instance and then bring it up again and make sure it still starts without issue.


migrations/schema.sql
```diff
@@ -12,23 +13,15 @@
...
-/*!50112 DEALLOCATE PREPARE s */;
 
 --
 -- Table structure for table `comments`
 --
 
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
 /*!40101 SET character_set_client = utf8 */;
-CREATE TABLE `comments` (
-  `repo` varchar(255) DEFAULT NULL,
```
copperwall: I'm not sure how this was `DEFAULT NULL`, primary key columsn can't be `NULL`.

entrypoint.sh
```diff
@@ -3,10 +3,12 @@
 
 # Wait for the DB to start accepting connections.
 for i in $(seq 5); do
+   sleep 10
    echo "Attempt $i"
-   node migrate.js && break || sleep 10
+   bin/migrate && break
 done
```
djmetzle: Can you use `until mysqladmin ping --silent; do sleep 1; done`?

djmetzle:
CR 👍

lexahall:
I was able to bring up a pulldasher instance using `docker-compose up`. I then killed this instance and brought it up again without any issues. QA 😸  but deploy_block :hand: incase it was wrong to use `docker-compose up` instead of just `docker-compose` (which didn't work for me).

entrypoint.sh
copperwall: I need to double check that the node:latest image has a MySQL client installed. I think I originally went with this because it didn't.

bin/migrate
```diff
@@ -13,7 +15,7 @@ var db = mysql.createConnection({
 console.log("Running this script will initalize your database for pulldasher.");
 console.log("It will not drop any existing tables.");
 
-fs.readFile('./schema.sql', function(err, data) {
+fs.readFile('migrations/schema.sql', function(err, data) {
```
addison-grant: Does this need to be `../migrations/schema.sql`?

bin/dump-schema
```diff
@@ -2,6 +2,7 @@
 
 # Executes mysqldump, passing through all arguments to this script.
 # invoke like dump-schema db_name -u ...
+SCHEMA_FILE="migrations/schema.sql"
```
addison-grant: Should this be `../migrations/schema.sql`?

entrypoint.sh
djmetzle: Ah. Okay then.

bin/migrate
copperwall: I thought so originally, but the relative path comes from where the script is executed, which is the root directory. Which is kind of confusing because the `require('../config')` is relative to where the actual migrate script is located, but the fs.readFile path isn't.

addison-grant: ohh

copperwall:
@lexahall `docker-compose up` is the right command to bring it up, sorry for not specifying that.

un_deploy_block :star:

bin/dump-schema
```diff
@@ -13,4 +14,6 @@ mysqldump "$@" \
    pull_labels \
    pull_signatures \
    pulls \
-   > schema.sql
+   > $SCHEMA_FILE
+
+sed -i -e "s/CREATE TABLE/CREATE TABLE IF NOT EXISTS/" $SCHEMA_FILE
```
danielj41: Huh, I guess there [isn't a mysqldump flag to do this](https://stackoverflow.com/questions/2220469/create-table-if-not-exists-from-mysqldump).

danielj41:
CR 👍

bin/dump-schema
copperwall: Yeah I tried to find one :(

Issue #87, opened 2017-08-07
RFC: Dealing with pulls that are not ready for QA


evannoronha:
We're running into circumstances where a pull ends up in the QA column even when it's not ready for QA. This happens in a couple different situations:
- Pull needs changes merged in
- Pull needs approval from ops
- QA needs to be done by a specific person outside QA/coders (i.e. Jeff Stephens QAing Admin pages)
- QA instructions are missing or don't fully test the new code path

This is problematic for QA because it keeps pulls in the QA column even though they fall into one of the above categories and cannot be QA'd. As we add more testers, it becomes more inefficient to have every tester look over a pull that they don't have the tools to sign off on, and it would be nice to draw attention to the pulls that need work or comments.

Proposed solution: Add a qa_block flag to Pulldasher. This flag would be added when a pull shouldn't be in the pool of pulls for testers to pick from. The pull would stay in the QA column on Pulldasher, but be moved to the bottom of the queue with a visual reminder (box shading or an icon) for testers not to open and for authors to investigate further.

Thoughts?

@iFixit/coders

sterlinghirsh:
Why not just dev_block it?

BaseInfinity:
Hmm all good points, I think we might be able to get away with using the `dev_block` label for things that are blocked by the developer

lexahall:
This is something that I've wanted since I started working here, but didn't think was completely necessary when we only had 1-2 active testers. Now that we've expanded to 5 testers, I think adding a qa_block would help to streamline the QA process and could significantly increase our efficiency. I spend a lot of time looking at same pulls over and over again, only to see that they still can't be QA'd yet. Additionally, I think it would be nice for pull owners to have a visual representation of which of their pulls are just waiting on QA and which are still in need of their attention before QA can begin. 

I'm open to different ideas on how to actually implement this, but I think @evannoronha suggestion is good.

djmetzle:
+1 for a `qa_block` label.

lexahall:
> Why not just dev_block it?
> Hmm all good points, I think we might be able to get away with using the dev_block label for things that are blocked by the developer

I don't think that it makes sense to dev_block in the cases that Evan and I are talking about. 
For one, dev_blocking a pull removes it from the CR column, which really sucks if it just needs more instructions added for QA. I don't want to hold a pull up and prevent it from getting CR just because it isn't quite ready for QA. 
[...]

PR #101, opened 2017-08-26
Add external block icon

### LINKED ISSUE #99: Add external block icon to pulldasher
In https://github.com/iFixit/pulldasher/issues/87 we decided to add an `external_block` label to pulls that require input or QA from someone outside of our department. I created the `external_block` label:
image 
But now we need to implement the icon in pulldasher. 

I was thinking we would use a red eye with a slash. Like so:
image
image

This mimics the QAing eye that we already have, which is green if someone else is QAing and blue if the viewer is QAing. I made it red with a slash to show that QA (by someone on the QA squad) is currently not possible.

lexahall:
In https://github.com/iFixit/pulldasher/issues/87 we decided to add an `external_block` label to pulls that require input or QA from someone outside of our department. When you add the `external_block` label to a github issue a red eye with a slash with show up next to the pull in the QAing column.
image

### QA
Come talk to me and we can QA this together. Or I can help get you set up with docker and explain how to QA it on your own. 

Closes #99

### DIFF
--- a/views/standard/spec/columns.js
+++ b/views/standard/spec/columns.js
@@ -293,6 +293,13 @@ define(['jquery', 'appearanceUtils', 'underscore'], function($, utils, _) {
                    label.created_at, label.user);
                   node.append(icon);
                }
+               if ((label = pull.getLabel('external_block'))) {
+                  var icon = $('<i>').addClass('fa fa-eye-slash externally-blocked');
+
+                  utils.addActionTooltip(icon, 'Externally Blocked',
+                   label.created_at, label.user);
+                  node.append(icon);
+               }
             }
          }
       }
--- a/views/standard/less/components/pull.less
+++ b/views/standard/less/components/pull.less
@@ -191,6 +191,12 @@
          color: @tag-qaing-mine;
       }
    }
+
+   &.externally-blocked {
+      color: @tag-externally-blocked;
+      border-color: @tag-externally-blocked-border;
+      background: @tag-externally-blocked-background;
+   }
 }
 
 .status {
--- a/views/standard/less/themes/colors/day_colors.less
+++ b/views/standard/less/themes/colors/day_colors.less
@@ -110,6 +110,10 @@
 @tag-qaing-border: @tag-default-border;
 @tag-qaing-background: @tag-default-background;
 
+@tag-externally-blocked: @red;
+@tag-externally-blocked-border: @tag-default-border;
+@tag-externally-blocked-background: @tag-default-background;
+
 @user-icon: @blue;
 
 // CR and QA boxes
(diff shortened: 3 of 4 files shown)


views/standard/spec/columns.js
```diff
@@ -293,6 +293,13 @@ define(['jquery', 'appearanceUtils', 'underscore'], function($, utils, _) {
                    label.created_at, label.user);
                   node.append(icon);
                }
+               if ((label = pull.getLabel('external_block'))) {
+                  var icon = $('<i>').addClass('fa fa-eye-slash externally-blocked');
```
kimarre: Does the selector here actually work? Isn't `$('')` just for id selectors? 
I don't think the using angle brackets within that is valid. I think `$E('i')` might be what was intended here, but it's probably even better to select it by a specific class name in case there are multiple `i`s.

views/standard/spec/columns.js
```diff
@@ -293,6 +293,13 @@ define(['jquery', 'appearanceUtils', 'underscore'], function($, utils, _) {
                    label.created_at, label.user);
                   node.append(icon);
                }
+               if ((label = pull.getLabel('external_block'))) {
```
colebemis: Is this supposed to be an assignment or a comparison?

views/standard/spec/columns.js
colebemis: I'm not familiar with this syntax `$('<i>')`. What is it supposed to select?

kimarre: I thought it was invalid too, but I guess it's not a selector, rather creates a new `i` element and applies it to the pull later.

colebemis: Whoa, ok. TIL

colebemis:
CR 👁

views/standard/spec/columns.js
kimarre: Assignment. It's using attributes of `label` a little lower down.

kimarre:
CR 💐

views/standard/spec/columns.js
lexahall: Assignment. Well, sort of both, but the single equals is right here. `label` gets assigned the result from  `pull.getLabel()`. If the label is `external_block` then the statement will evaluate to true, and you'll enter the if.

views/standard/spec/columns.js
lexahall: Since the icon we want doesn't exist yet, we can't just select it. This creates a new `<i>` tag and then adds the appropriate classes to it. Then we use `node.append(icon)` to place it where we want.

views/standard/spec/columns.js
addison-grant: It looks like there's an extra set of parens. Is that intentional?

mlahargou:
QA 📱  We tested out multiple cases and they worked. 👍

views/standard/spec/columns.js
lexahall: I just stayed consistent with the if for the QAing label. I think the parens could probably be removed. Want me to take them off both?

Issue #93, opened 2017-08-11
CI: Bypass the CI column if there's no integrations for a repo / pull


kwiens:
![image](https://user-images.githubusercontent.com/129007/29223734-7bcb1786-7e7c-11e7-8d34-06371a7994f6.png)

copperwall:
https://github.com/iFixit/pulldasher/blob/master/views/standard/spec/columns.js#L55

The `selector` function for the CI Blocked column is a good place to start.

danielbeardsley:
We can't automatically know if there are any integrations, we just know that CI hasn't run for this pull. The commit status is something that's stored per-commit, and its absence doesn't tell you anything about the future (maybe CI hasn't run, maybe CI will never run, ...)

I'd lean toward making this configurable per-repo in the config: `ci_required: false` or something.

copperwall:
This might be a good reason to have the `repos` list in `config.js` be an object instead of a `org/repo` string.

copperwall:
This was addressed in the short term with https://github.com/iFixit/pulldasher/issues/130 and is probably going to be long term fixed by #102

Issue #72, opened 2017-02-25
CI Builds: Support multiple


copperwall:
We currently only support one CI status per commit/pull. Github lets you do multiple and we have multiple checks for the time being.

It'd be cool to see a tool tip saying `x/y builds passed` on the CI status bar.

PR #102, opened 2017-08-27
Multiple configurable build statuses

g1y:
Allow multiple configurable build statuses for passing builds in pulldasher.

#### Problems this solves:

Right now in pulldasher if a build doesn't have a build at all, it will be failed, because there is no "passing".

Along with that, if we have multiple build statuses, pulldasher only looks at the first one.

To solve both of these problems, required builds can be specified per-repo, and a pull will only be passing if it has a successful status for each of these.

#### Expected behaviour:

- Adding a required build to a repo that doesn't have any builds should not be passing
- Adding a required build to repo that has a build that is passing and is the required context should pass if it wasn't the same context it should not pass.
- Having any one required build not be in the passing builds should cause it not to pass.
- If there are extra builds, but not required, they shouldn't affect whether this is passing or not.

#### QA
- Assert expected behaviours listed above for a single build.
- Assert expected behaviours listed above for multiple builds with one required and others optional.
- Assert expected behaviours listed above for for multiple builds statuses all required.

Closes: https://github.com/iFixit/pulldasher/issues/93
Closes: https://github.com/iFixit/pulldasher/issues/72


danielbeardsley:
> To solve both of these problems, required builds can be specified per-repo, and a pull will only be passing if it has a successful status for each of these.

That's rad!!

MooreGuy:
dev_block :+1: I'll finish this up next weekend.

lib/utils.js
```diff
@@ -41,9 +41,16 @@ module.exports = {
...
+
+      return Promise.all(this.getRepoFullNames().map(allRepoMap))
       .then(function(repoItems) {
          return _.flatten(repoItems, /* shallow */ true);
       });
-   }
+   },
+
+   getRepoFullNames: function() {
+      return config.repos.map(function(repo) {
```
copperwall: Now that the dev environment has node v6 this can be

```js
return config.repos.maps(repo => repo.full_name);
```

:D

models/pull.js
```diff
@@ -240,4 +241,22 @@ Pull.getFromDB = function(data, signatures, comments, commitStatus, labels) {
...
 };
 
+Pull.isPassing = function() {
+   var required_builds = config.repos.find(function(repo) {
+      if (repo.full_name === this.repo) {
+         return repo.required_builds;
+      }
+   }).required_builds;
+
+   var passing_builds = this.commitStatuses.reduce(function(passingAccum, build) {
```
copperwall: Instead of a reduce, this can just be a filter for statuses that have a `state` of "success"

config.example.js
```diff
@@ -57,7 +57,12 @@ module.exports = {
 
    // List of repositories for pulldasher to watch.
    repos: [
-     "owner/name"
+      {
+         full_name: "owner/name",
```
copperwall: I don't really know how github describes the full name of a repo, so maybe this isn't helpful, but would it be better to store the org name and repo name as their own fields in the object and then have `getRepoFullNames` concat those individual values into the canonical `org/repo` string?

I can't think of a use case now, but if we ever wanted to get an org name out of a repo object, I think it'd be easier as a field access than having to parse out the name from `full_name`.

models/pull.js
```diff
@@ -240,4 +241,22 @@ Pull.getFromDB = function(data, signatures, comments, commitStatus, labels) {
    return new Pull(pullData, signatures, comments, commitStatus, labels);
 };
 
+Pull.isPassing = function() {
+   var required_builds = config.repos.find(function(repo) {
+      if (repo.full_name === this.repo) {
+         return repo.required_builds;
```
copperwall: Should the find callback just return whether or not the repo.full_name matches the repo name?

lib/utils.js
MooreGuy: Planning on doing this

config.example.js
MooreGuy: I agree, it will probably come into use. The github object for it holds the full_name and the repo, owner all in separate variables.

I think this is totally reasonable to add when it is desired and could be simply implemented as parsing the full_name from the config.

lib/utils.js
MooreGuy: 5deb403

models/pull.js
MooreGuy: c5fad924260069eb57e256e87019118a8f49e48c

models/pull.js
MooreGuy: 79eb6e6

MooreGuy:
un_dev_block :+1: ready!

models/pull.js
```diff
@@ -28,6 +28,18 @@ function Pull(data, signatures, comments, commitStatus, labels) {
       this.data.closes = data.closes;
       this.data.connects = data.connects;
    }
+
+   this.isPassing = function() {
+      var repoConfig = config.repos.find(repo => repo.full_name === this.data.repo);
+      var requiredBuilds = repoConfig.required_passing_builds;
```
danielj41: This is neat!

views/standard/spec/columns.js
```diff
@@ -229,7 +229,7 @@ define(['jquery', 'appearanceUtils', 'underscore'], function($, utils, _) {
          id: "qaPulls",
          selector: function(pull) {
             return !pull.qa_done() && !pull.dev_blocked() &&
-             pull.build_succeeded();
+             pull.passing;
```
danielj41: Any particular reason for using `passing` rather than `isPassing()`? It seems safer to always use a computed value.

danielj41:
CR 👍

views/standard/spec/columns.js
MooreGuy: This is on the frontend, where we don't have access to the config. ideally, yes if we didn't have this constraint, I would have preferred it to do it the way you suggested.

danielj41: Makes sense. I didn't realize this ran on the frontend.

public/js/Pull.js
```diff
@@ -100,19 +100,8 @@ define(['underscore', 'appearanceUtils', 'socket'], function(_, utils, socket) {
...
-
-      build_failed: function() {
-         var status = this.build_status();
-         return status === 'failure' || status === 'error';
-      },
-
-      build_succeeded: function() {
-         return this.build_status() === 'success';
-      },
+      build_failed: () => !this.passing,
```
copperwall: Nothing to block on, but in ES6 you can do

```js
build_failed() {
   return this.passing;
}
```

but it's basically the same.

I'm not sure what the convention is for having `this` bound functions as object methods. I guess it only matters if you're planning on using `call` or `apply`, but maybe that's bad practice.

models/pull.js
```diff
@@ -28,6 +28,18 @@ function Pull(data, signatures, comments, commitStatus, labels) {
       this.data.closes = data.closes;
       this.data.connects = data.connects;
    }
+
+   this.isPassing = function() {
```
copperwall: If this is only really meant to be called during the constructor, what do you think about making it private? If I saw that the Pull object had both a `passing` field and an `isPassing` method I don't think it'd be immediately obvious as to which one is the canonical representation of whether or not a pull is passing.

You could keep the name, but just define it within the scope of the `pull` module without placing it on the actual Pull object.

MooreGuy: Yes, it only needs to be calculated when build statuses change, which they never do other than on construction.

I'm game for making it private, I agree it would make it much more clear which one should be used.

public/js/Pull.js
MooreGuy: Got removed later, I felt like it was silly to have one method that just negated the other.

MooreGuy:
dev_block :+1: on copperwall's suggestion

MooreGuy:
un_dev_block :+1:

danielj41:
CR 👍

copperwall:
CR :fire:

addison-grant:
Would you like some QA on this? I am not sure how to configure different CI builds so we can have multiple builds on a repo.

lexahall:
@MooreGuy bump
What should QA look like?

evannoronha:
dev_block 🔥 on QA tips.

MooreGuy:
Adding multiple builds on a repo can be done with cimpler, and travis ci together. If you add a cimpler build status and then add a travis build you'll have two. To specify which ones are required for passing, change the config to match this format, https://github.com/iFixit/pulldasher/pull/102/files#diff-02880f48501a3eb52890d5181ba67496R62

Clicking on the build status pull out bar will always go to just one build. This should be ignored, it's not fully decided how to correctly display multiple build statuses.

lib/git-manager.js
```diff
@@ -412,17 +411,13 @@ function getCommit(repo, sha) {
...
-      api.getCommitStatus(params({
+      api.getCommitStatuses(params({
          sha: sha
       }, repo)).then(function(status) {
-         if (status.length) {
-            resolve(status[0]);
-         } else {
-            resolve(null);
-         }
+         resolve(status);
```
danielbeardsley: statuses?

lib/db-manager.js
```diff
@@ -377,14 +377,14 @@ var dbManager = module.exports = {
...
+   getCommitStatuses: function(repo, sha) {
       var q_select = '\
          SELECT * FROM commit_statuses \
          WHERE commit = ? AND repo = ?';
 
       return db.query(q_select, [sha, repo]).then(function(rows) {
-         if (rows.length === 1) {
-            return Status.getFromDB(rows[0]);
+         if (rows.length > 0) {
+            return rows.map(row => Status.getFromDB(row));
```
danielbeardsley: Should this perhaps return an empty array instead of null when now rows are found? 

Looks like it's eventually ORed to [an empty array anyway](https://github.com/iFixit/pulldasher/blob/556979983ed2eca064ca90c6ce104caf122754b8/models/pull.js#L14) so I think it makes sense to do that here so the type is consistent.

models/pull.js
```diff
@@ -28,6 +28,18 @@ function Pull(data, signatures, comments, commitStatus, labels) {
...
    }
+
+   var isPassing = function() {
+      var repoConfig = config.repos.find(repo => repo.full_name === this.data.repo);
+      var requiredBuilds = repoConfig.required_passing_builds;
+
+      var passingBuilds = this.commitStatuses.filter(
+         build => build.state === 'success');
+
+      return !!_.without(requiredBuilds, ...passingBuilds);
```
danielbeardsley: There's a bug here. requiredBulds is an array of strings. passingBuilds is an array of commit status objects.

I think this would be better expressed like:
```
required_builds.all(required_build -> 
   passingStatuses.find(status -> status.branch == required_build)
)
```

views/standard/spec/indicators.js
```diff
@@ -214,8 +214,8 @@ define(['jquery', 'underscore', 'appearanceUtils', 'debug'], function($, _, util
          signatureStatus(pull, node, 'QA', required, pull.qa_signatures);
       },
       build_status: function status(pull, node) {
-         if (pull.status.commit_status) {
-            var commit_status = pull.status.commit_status.data;
+         if (pull.status.commit_statuses) {
+            var commit_status = pull.status.commit_statuses[0].data;
```
danielbeardsley: We're picking an arbitrary one. It may not even be one of the required ones. Perhaps a specific one would be better, or perhaps the first one in the required list?

Oh, maybe this pull was more about dealing with the backend? Leaving the front-end changes for a future pull?

models/pull.js
```diff
@@ -28,6 +28,18 @@ function Pull(data, signatures, comments, commitStatus, labels) {
...
+   };
+
+   this.data.passing = isPassing();
```
danielbeardsley: I don't feel great about this. I'd much rather leave this calculation dynamic than effectively "cache" it as a column in the table.

They each have their advantages. Maybe we should do both? Having it "cached" in the DB certainly makes it easier for analysis and dashboards in SQL. Having it calculated makes it better for dealing with changes in the config, we don't have to clear or recalculate this column.

Oh, this is recalculated each time the pull is loaded from the DB (which is every time it's modified / commented / ...) So that's often enough.

danielbeardsley:
dev_block 👍  on at least the bug.

lib/git-manager.js
g1y: 👍

g1y: 1c584f19f2d102b

views/standard/spec/indicators.js
g1y: e3e75488efa9428d121047d16b2a645ac5f8c639

models/pull.js
g1y: 1e9bb376676a6ad3be9ddd32622cc34700e8ed7d

lib/db-manager.js
g1y: e3e75488efa9428d121047d16b2a645ac5f8c639

views/standard/spec/indicators.js
g1y: Yeah, that's another thing. There isn't an obvious answer to doing this, so it'll have to be done in a following pull.

g1y:
un_dev_block :+1: Check that master merge carefully. It was not easy to read through it.

addison-grant:
Some of these are difficult to understand:

> - Adding a required build to a repo that doesn't have any builds should not be passing

I think this is what you mean, "If there is a pull without any builds in a repo with a required build, then that pull should not be passing."

> - Adding a required build to repo that has a build that is passing and is the required context should pass if it wasn't the same context it should not pass.

I think you mean: "If there is a pull with passing builds, but a new required build is added, then the pull should not be counted as passing until the newly required build and all the other builds on that pull are passing."

> - Having any one required build not be in the passing builds should cause it not to pass.

Makes sense.

> - If there are extra builds, but not required, they shouldn't affect whether this is passing or not.

Makes sense.

lib/git-manager.js
```diff
@@ -410,17 +409,13 @@ function getCommit(repo, sha) {
    });
 }
 
-function getCommitStatus(repo, sha) {
+function getCommitStatuses(repo, sha) {
    return new Promise(function(resolve, reject) {
```
copperwall: It seems kind of weird that we're wrapping a promise in our own promise instead of just passing the `api.getCommitStatuses` promise.

lib/db-manager.js
```diff
@@ -38,9 +38,9 @@ var dbManager = module.exports = {
       return dbManager.updatePull(pull).then(function() {
          var updates = [];
 
-         if (pull.commitStatus) {
+         if (pull.commitStatuses) {
             updates.push(
-               dbManager.updateCommitStatus(pull.commitStatus)
+               pull.commitStatuses.forEach(status => dbManager.updateCommitStatus(status))
```
danielbeardsley: `forEach()` doesn't return a promise... it doesn't return anything I think. Maybe you mean `.map()` ? and then a splat?

lib/git-manager.js
danielbeardsley: I agree. We're just passing it through here and losing any rejections.

danielbeardsley:
CR 👍 

But dev_block 👍  on a few small things.

lib/db-manager.js
g1y: Yeah push call should have been moved in the forEach

lib/db-manager.js
```diff
@@ -376,17 +376,13 @@ var dbManager = module.exports = {
...
          SELECT * FROM commit_statuses \
          WHERE commit = ? AND repo = ?';
 
       return db.query(q_select, [sha, repo]).then(function(rows) {
-         if (rows.length === 1) {
-            return Status.getFromDB(rows[0]);
-         } else {
-            return null;
-         }
+         return rows.map(row => Status.getFromDB(row));
```
copperwall: oowee

BaseInfinity:
Is this ready for CR?

models/pull.js
```diff
@@ -28,6 +28,19 @@ function Pull(data, signatures, comments, commitStatus, labels) {
       this.data.closes = data.closes;
       this.data.connects = data.connects;
    }
+
+   var isPassing = function() {
```
copperwall: Could've been a short function to get implicit this binding, but eh.

Also now that we're running a sorta recent version of node we probably shouldn't be too shy about using `const` instead of `var`.

models/pull.js
```diff
@@ -28,6 +28,19 @@ function Pull(data, signatures, comments, commitStatus, labels) {
...
+      var repoConfig = config.repos.find(repo => repo.full_name === this.data.repo);
+      var requiredBuilds = repoConfig.required_passing_builds;
+
+      var passingBuilds = this.commitStatuses.filter(
+         build => build.state === 'success')
+         .map(build => {return build.data.context;});
```
copperwall: This can just be `.map(build => build.data.context)`

copperwall:
CR 🔥

models/pull.js
g1y: right-o. When you're still not comfortable with those ES6 features :cry:

BaseInfinity:
CR 👙  would love to see this merged

g1y:
I had a bit of trouble trying to test this out, I'll try again.

danielj41:
CR :trollface:

g1y:
builds aren't showing up in the UI, dev_block :+1: on that

g1y:
So here is my QA list

Make a pull with a single required build
- [ ] if it fails it should be in the failing column
- [ ] if it is passing it should not be in the passing column
- [ ] When it gets full review and is passing it should show up in the ready
- [ ] When it gets full review but is failing it shoudl show up in just the CI blocked column
Make a second pull with two required builds
- [ ] if one is failing, and one is passing it should be in CI blocked
- [ ] if both are failing, then it should be in the CI blocked
- [ ] if both are passing, then it should be in the ready if reviewed, otherwise show up in the CR or QA columns
Make a pull with no required builds but add a CI status to it anyway
- [ ] failling build should still be passing
- [ ] passing build should be passing

Double check for regressions:
- [ ] CR's should show up on a pull
- [ ] QA's should show up on a pull
- [ ] dev_blocks should move a pull to the dev_block column
- [ ] changing cr_req should properly affect the CR requirements

Issue #106, opened 2017-09-18
Webhooks: A few events causing pulldasher to crash


g1y:
There are a few events that are causing pulldasher to crash:
```
{
  "action": "milestoned",
  "issue": {
...
```

Gives a stack trace of: 
```
Sep 18 16:31:15 pulldasher[8365]: TypeError: Cannot read property 'base' of undefined
Sep 18 16:31:15 pulldasher[8365]:     at Function.Pull.fromGithubApi (/opt/pulldasher/models/pull.js:165:17)
Sep 18 16:31:15 pulldasher[8365]:     at /opt/pulldasher/controllers/githubHooks.js:160:43
Sep 18 16:31:15 pulldasher[8365]:     at /opt/pulldasher/node_modules/promise/index.js:18:19
Sep 18 16:31:15 pulldasher[8365]:     at flush (/opt/pulldasher/node_modules/asap/asap.js:27:13)
Sep 18 16:31:15 pulldasher[8365]:     at _combinedTickCallback (internal/process/next_tick.js:73:7)
Sep 18 16:31:15 pulldasher[8365]:     at process._tickCallback (internal/process/next_tick.js:104:9)
```

#### Problem:
Strange event types cause pulldasher to crash, if we encounter a weird event, it should be skipped over.

PR #58, opened 2017-01-06
Hooks: deal with issue events about pull requests

danielbeardsley:
We've seen this in practice and the docs mention it; we can receive
issue events about pull requests. Thankfully they also mention how to
distiguish them. Previously, we didn't and would occasionally get
issues records created for pull requests (instead of records in the
pulls table).

CC @entendu

Note: there are a few other small refactorings in here too, go
commit by commit for CR.

Closes #57

PR #110, opened 2017-09-22
Stop trying to update pull on issue events

MooreGuy:
There are separate events for pulls, and I can't think of any case where
an issue event needs to have a pull updated.

QA:

Replay the previous webhooks that failed, or get github to fire those webhooks by manually doing those operations. Issues events should no longer make pulldasher restart.

### DIFF
--- a/controllers/githubHooks.js
+++ b/controllers/githubHooks.js
@@ -130,8 +130,6 @@ var HooksController = {
 
 function handleIssueEvent(body) {
    debug('Webhook action: %s for issue #%s', body.action, body.issue.number);
-   // Some of these events are also triggered on pull requests
-   var isPull = body.issue.pull_request;
 
    var doneHandling = handleLabelEvents(body);
 
@@ -151,14 +149,6 @@ function handleIssueEvent(body) {
       case "assigned":
       case "unassigned":
         // Default case is update the issue
-   }
-
-   // If github is fooling us and this is actually a pull request,
-   // let's not create an issue object out of it.
-   if (isPull) {
-      return doneHandling.then(function() {
-         return dbManager.updatePull(Pull.fromGithubApi(body.pull_request));
-      });
    }
 
    return doneHandling.then(function() {


danielj41:
This is the commit that introduced that code: 4711d6b2bb3d9a23b2bf248edeac49dba88c6ce2

> Hooks: deal with issue events about pull requests
> 
> We've seen this in practice and the docs mention it; we can receive
issue events about pull requests. Thankfully they also mention how to
distiguish them. Previously, we didn't and would occasionally get
issues records created for pull requests (instead of records in the
pulls table).

Do the docs still mention this? Maybe it got fixed by github?

MooreGuy:
I'll surf around for them

danielj41:
CR 👍

MooreGuy:
Good call, I didn't find anything in there: https://developer.github.com/v3/activity/events/types/#issuesevent

BaseInfinity:
CR 👙

lexahall:
What should QA look like on this?

davidrans:
The docs still say the same thing: https://developer.github.com/v3/issues/#list-issues

lexahall:
QA 😸

danielbeardsley:
Possible that we traded one problem for another.
This pull was an attempt to fix #106 

It re-introduces the problem fixed by #58 which will cause this problem to grow:
```
mysql> select count(*) from pulls join issues using (number, repo);
+----------+
| count(*) |
+----------+
|      147 |
+----------+
1 row in set (0.04 sec)
```

It seems like the data coming in on issue events about pull requests looks different than it used to... or we are expecting different fields than we used to.

Issue events about pull requests now have (or maybe always have?) a body that looks just like an issue, i.e. an issue event about an issue vs a pull have identical structure and thus only a subset of the pull fields. So, I don't think we should update / create a row in the `issues` table for a pull request (like the code now does). I think our code should have been adjusted to deal with missing data instead.

MooreGuy:
asdf

PR #115, opened 2017-10-03
Add migration system

g1y:
Adds an actual migration system so that migrations do not have to be run by hand.

Only adds the dependency and an example config file.

### DIFF
--- a/database.example.js
+++ b/database.example.js
@@ -1,0 +1,8 @@
+{
+   "prod": {
+      "driver": "mysql",
+      "user": "root",
+      "password": "root",
+      "multipleStatements": true
+   },
+}
--- a/lib/db-manager.js
+++ b/lib/db-manager.js
@@ -153,7 +153,7 @@ var dbManager = module.exports = {
     */
    insertSignature: function(signature) {
       debug('Calling insertSignature for pull #%s in repo %s',
-         signature.data.number, signature.repo);
+         signature.data.number, signature.data.repo);
       var dbSignature = new DBSignature(signature);
       var repo = dbSignature.data.repo;
       var number = dbSignature.data.number;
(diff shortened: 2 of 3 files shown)


database.example.js
```diff
@@ -0,0 +1,8 @@
+{
```
danielj41: I think this should be `database.example.json`: 

> Location of the database.json file.             [default: "./database.json"]

from https://db-migrate.readthedocs.io/en/latest/Getting%20Started/installation/

danielj41:
I think it'd be useful to include a sample migration that doesn't do anything. If nothing else, it'd help QA this pull.

danielj41:
CR 👍  but dev_block 👍  on the `json` config file.

database.example.js
MooreGuy: :+1: Ah, yes, thank you.

MooreGuy:
Yeah, I that's probably a good idea, to show that it works, I'll think of a migration to add.

danielbeardsley:
> I'll think of a migration to add.

You could make a no-op migration to show how it works.

I read through all the docs of `db-migrate` and while it sounds like a cool tool, they don't really explain how to go about start using it with an existing project. i.e. should we use our entire schema.sql file as the first migration?

I just don't see us having enough migrations to make this worthwhile, but maybe I'm not thinking far enough into the future?

sterlinghirsh:
Status update? This has been blocked for months.

g1y:
@copperwall was looking into a nice RDBMS, but I 'm not sure if he'll be going forward with that. If he did decide he wanted to use that, this would definitely conflict with that. I want his comment on what he wants to do before going forward with this.

Otherwise, I'm sure people would love to have this.

copperwall:
I still want to do that D: We can talk about it more when I get back next week

PR #116, opened 2017-10-03
Use new org/repo string for Cryogenic pull links

copperwall:
We were using a leftover `org` variable to construct a string to the
github page where all pulls with the cryo label exist. The new way a
repo is represented is a string like "org/repo", so instead of relying
on an org and a repo, we can just include that string as the first two
parts of the github URL path.


MooreGuy:
CR :+1:  danke schön

PR #118, opened 2017-10-10
Labels: Remove extra repo name parameter

### LINKED ISSUE #117: QAing Eye: Does not have an associated user
The QAing eye and externally blocked eye are borked. They look like this:
image

image

It seems like this is a webhook bug because Pulldasher is not moving QA pulls with my label to the top, nor using the blue eye to indicate ownership.

CC @MooreGuy

copperwall:
The `Label` constructor and its callsites were using different numbers of
parameters. For the few that were using too little, the fields were
being incorrectly set to the wrong label fields, i.e. the repo field
would be set to the user value.

This removes the extra constructor parameter and removes arguments left
over from before the `repo_name` column was removed.

qa_req 0

screenshot from 2017-10-09 22-13-30

This also works for label add and delete webhooks.

Closes #117

### DIFF
--- a/controllers/githubHooks.js
+++ b/controllers/githubHooks.js
@@ -177,7 +177,6 @@ function handleLabelEvents(body) {
          return dbManager.insertLabel(new Label(
             body.label,
             object.number,
-            body.repository.name,
             body.repository.full_name,
             getLogin(body.sender),
             object.updated_at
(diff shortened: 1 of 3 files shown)


BaseInfinity:
CR 👙

danielj41:
CR 👍

PR #121, opened 2017-10-14
Start node inspector when DEBUG is set

copperwall:
When the DEBUG environment variable is set, make sure to start
Pulldasher with the `--inspect` flag. This lets you connect a debugger
remotely (DevTools for Node, node-inspector).

cr_req 1
qa_req 0 I started up a debug instance of pulldasher with Chrome DevTools open and successfully attached.


copperwall:
CR :guy looked at it:

PR #123, opened 2017-10-16
Webpack: Bundle JS with webpack

### LINKED ISSUE #97: Replace bower with npm (and webpack?)
Bower tells you not to use it when you install it, so maybe we should listen.

Bower is currently responsible for installing the following front end modules:
* bootstrap
* font-awesome
* requirejs
* requirejs-text
* visionmedia/debug

Our less files expect to be able to import the asset files directly from`bower_components`, so we need to copy the files from `node_modules` to a public directory.

## Bonus

Can we replace requirejs with webpack without changing all of the the `require` and `define` loaders in the front end files?

This blog post might be a good resource for doing that https://medium.com/@andrejsabrickis/bye-bye-bower-or-how-to-migrate-from-bower-to-npm-and-webpack-4eb2e1121a50

copperwall:
This replaces require.js and bower with webpack and npm for bundling and loading front end JavaScript files. By using webpack, this makes it easy for us to make use of loaders to incorporate things like babel, uglify, eslint, and maybe even typescript into our build process.

This pull does these main things:
* All uses of `define(['jquery', 'underscore'], function($, _) {`, etc. are replaced with ES6 import statements.
* All modules handled by bower are now handled by npm

### CR Notes

A huge amount of the diff is just whitespace/tabs. `?w=1` is your friend.

### QA Notes

~~This should work the same as `master` does. I'd recommend going with the docker-compose route to make sure that you can spin up a local dev environment and have it work off the bat.~~

This has been sitting/being used in staging for over a day and nothing crazy has happened, so I'm inclined to just merge it.

qa_req 0

Closes #97


Dockerfile
```diff
@@ -4,12 +4,9 @@ RUN mkdir -p /opt/pulldasher
 WORKDIR /opt/pulldasher
 
 # Install dependencies
-COPY package.json /opt/pulldasher
-RUN npm install
-
 COPY . /opt/pulldasher
+RUN npm install --unsafe-perm
```
danielj41: Instead of adding `--unsafe-perm`, can we just not run `npm install` as root?

I may be misunderstanding the situation.

views/layout.html
```diff
@@ -9,31 +9,15 @@
...
-            },
-            'debug': {
-               exports: 'debug'
-            }
-         }
-      };
+         var App = {
+            socketToken: "<%= socketToken %>",
+            user: "<%= user %>",
+            debug: <%= debug %>
```
danielj41: Does this introduce any xss vulnerabilities?

danielj41: Never mind, this is moved from another file.

webpack.config.js
```diff
@@ -0,0 +1,31 @@
...
+            use: 'raw-loader'
+         }
+      ]
+   },
+   resolve: {
+      modules: ['public/js', 'views/current', 'node_modules']
+   },
+   plugins: [
+      new webpack.ProvidePlugin({
+        $: "jquery",
```
danielj41: Is this necessary? It looks like you have `import $ from 'jquery'` in most files.

danielj41:
This is great, thanks @copperwall!

CR 👍  with a few minor questions.

Dockerfile
copperwall: Hmmm, I think with the way the container is set up, root is the only user. I haven't added anything to create a new user just because the node app is the only thing running in the container. I'm not sure if that's an anti-pattern though.

I'll see if I can find documentation saying to do one or the other.

webpack.config.js
copperwall: Ah yeah, so this is necessary to make Bootstrap not die when bundling. I didn't realize this until after I put `import $ from 'jquery'` in our files that needed it, so I suppose I don't really need all those explicit jquery imports.

Is it common to just add utility libraries to ProvidePlugin so that there's not an extra `import _` or `import $` in every file, or is it better to be explicit?

danielj41: That makes sense. Thanks for explaining.

I think it's better to keep it explicit in our own source files.

copperwall: Cool! I think that sounds good. If we ever change out bootstrap for something else we could remove the provider plugin and then we won't have to re-add imports to all of our files.

views/layout.html
copperwall: It doesn't _introduce_ any ;)

I haven't tried though. Maybe it's better to get this information as JSON from the backend instead of templating it. I'm okay with adding another issue to look into changing how App is populated.

mlahargou:
dev_block 📱  on conflicts 👍

g1y:
CR :+1:

danielj41:
CR 👍

views/standard/spec/indicators.js
```diff
@@ -185,6 +185,7 @@ var signatureStatus = function(pull, node, type, required, signatures) {
       }
    };
 
+<<<<<<< HEAD
```
g1y: I think you missed a piece

g1y:
CR 👍 but dev_block 👍 on the conficts header still in there

views/standard/spec/indicators.js
copperwall: Oops, that explains the CI build failure

copperwall:
un_dev_block ⭐️

g1y:
boom CR :boom:

danielj41:
CR 👍

Dockerfile
danielbeardsley: I would say that inside a container, root isn't really an "elevated" privilege, right?

danielbeardsley:
CR 👍

PR #131, opened 2018-02-09
Shows pulls from repos without CI in QA column

### LINKED ISSUE #130: Items without CI are stuck in CI-blocked
## Issue

Pulls without CI set up in their repo are always CI blocked, requiring circumventing Pulldasher and getting QAs manually for any pulls in those repos.

## Do this

Make CI-less pulls not block in the CI Blocked column.

n-fisher:
This pull changes QA column behavior to include any pulls that don't have CI set up in their repo. Previous behavior would block these pulls from showing up outside of CI-Blocked and CR columns.

| Current | Updated |
|---------|---------|
| image | image |

CR
--
- `pull.status.commit_status` is an object that is falsey (`null`) if there's no CI set up, and truthy if there is CI but it's failing or not ran yet.

QA
--
Talk to me about how to set up a test instance of Pulldasher.

Closes #130

### DIFF
--- a/views/standard/spec/columns.js
+++ b/views/standard/spec/columns.js
@@ -230,7 +230,7 @@ export default [
       id: "qaPulls",
       selector: function(pull) {
          return !pull.qa_done() && !pull.dev_blocked() &&
-          pull.build_succeeded();
+          (pull.build_succeeded() || !pull.status.commit_status);
       },
       sort: function(pull) {
          // The higher score is, the lower the pull will be sorted.
--- a/views/standard/spec/columns.js
+++ b/views/standard/spec/columns.js
@@ -54,7 +54,8 @@ export default [
       // `false` otherwise. This property may also be an array of functions,
       // in which case the selectors are chained.
       selector: function(pull) {
-         return !pull.dev_blocked() && !pull.build_succeeded();
+         return !pull.dev_blocked() && !pull.build_succeeded() &&
+          !pull.build_unavailable();
       },
       // This describes the sort order for the pull. It returns a numeric
       // score for each pull. Pulls with a low score are sorted to the bottom
@@ -102,7 +103,8 @@ export default [
       title: "Deploy Blocked Pulls",
       id: "deployBlockPulls",
       selector: function(pull) {
-         return pull.ready() && pull.deploy_blocked();
+         return (pull.ready() && pull.deploy_blocked()) ||
+          pull.buildUnavailableReady();
       },
       triggers: {
          onCreate: function(blob) {
@@ -230,7 +232,7 @@ export default [
       id: "qaPulls",
       selector: function(pull) {
          return !pull.qa_done() && !pull.dev_blocked() &&
-          (pull.build_succeeded() || !pull.status.commit_status);
+          (pull.build_succeeded() || pull.build_unavailable());
       },
       sort: function(pull) {
          // The higher score is, the lower the pull will be sorted.
--- a/views/standard/spec/columns.js
+++ b/views/standard/spec/columns.js
@@ -126,6 +126,12 @@ export default [
                current_block.created_at, current_block.user.login);
 
                node.append(link);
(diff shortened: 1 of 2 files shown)


views/standard/spec/columns.js
```diff
@@ -230,7 +230,7 @@ export default [
       id: "qaPulls",
       selector: function(pull) {
          return !pull.qa_done() && !pull.dev_blocked() &&
-          pull.build_succeeded();
+          (pull.build_succeeded() || !pull.status.commit_status);
```
copperwall: This might cause a problem for pulls that haven't run in CI yet. When CI is particularly backed up, it may be some time before a commit even gets an in progress status. So things would catch pulls from repos that either have no CI, or do have CI but haven't been run through it yet.

@g1y was working on an implementation to add context to a repo in terms of what "passing" means in #102.

n-fisher: But the pulls get updated via webhook, right? So when CI unlocks (just queues the PR for CI, doesn't even need to start building/tests) it will update the PR back into the CI blocked column? 

In my opinion we have two options: 
1. Blocking QA on legitimate PRs
2. Risking an occurrence where CI is so backed up that a pull gets QA attention before CI catches up, and then the pull fails CI later.

It seems like 2 is rare (requires CI lag that lasts longer than a QA queue+QA session AND the pull subsequently fails CI) while 1 is something I've been burned by usually one or two pulls per sprint.

copperwall: That's true, CI backlog hasn't been too much of an issue recently. Something else to consider is how pulls from repos with no CI show up in the CR column and the Ready column. If we don't cover those bases too then there might be issues with visibility for reviewers. Or a pull could be completely signed off without showing up in the ready column, or otherwise communicating that it's good to go.

Following down that path, if we put pulls in the ready column that may or may not have CI, that might communicate a pull is good to go, when it's still waiting for CI to run on it at all.

It'd be nice if there was something in the Github API to tell us whether or not a repo has any continuous integration services associated with it. That way we wouldn't need manual configuration like #102 and we could be a little more confident about whether or not something actually has CI.

n-fisher: Agreed, and I wouldn't want CI-failling pulls to get deployed. That said, it's hard to imagine a scenario in which we have CI that queues more slowly than 2 crs and a QA. Also, if I'm reading it right the `build_succeeded` function is only true if the build status is explicitly "success", and the ready column requires `build_succeeded`. 

So I don't think we'd see it show up in the ready column either way in this pull, but will at least get to see it in QA.

n-fisher: What if we filter CI-less Ready pulls into the deploy-blocked column, to be deployed by someone who's aware that there's no CI? That way it will deploy-block by default, meaning we'd avoid CI-lag causing a pull to falsely be listed under Ready and at the same time avoid QA missing the pull. Plus, the author will expect it to not have CI and will watch Deploy-Blocked as their Ready column for CI-less pulls.

n-fisher:
dev_block :on deploy_blocking ready pulls without CI:

n-fisher:
un_dev_block 👍

k0rvusk0r4x:
Worked with @n-fisher on this:

QA 😺 Everything is working nicely.

dev_block 😿 to give an icon indicating why the pull is deploy blocked. ("CI not required"/"CI not found"/etc.)

lexahall:
CR :octocat:

n-fisher:
boom.

image

n-fisher:
un_dev_block 👍

lexahall:
CR :octocat:

k0rvusk0r4x:
Everything is still looking great!

QA 😺

copperwall:
CR :LGTM:

danielbeardsley:
> pulls that don't have CI set up in their repo

The problem is we don't have this data. We only know that CI hasn't yet run for the given pull, we don't know if it's supposed to.

PR #135, opened 2018-02-27
Use bluebird promises

### LINKED ISSUE #76: Use ES6 Promises
We do a lot of `Promise.denodeify` in our Github library wrapper because the `github` module didn't support Promises at the time. Now it does, so after we update that usage, we won't have a lot of use for `Promise.denodeify` and won't need an extra Promise library.

copperwall:
Bluebird is a pretty performant promise library that implements support
for things like .done and Promise.promisfy that we currently use in
Pulldasher. It's written in a way that's supposed to use less memory
than standard ES6 promises, so I thought it'd be interesting to try it
out.

From some light testing, running pulldasher with then/promise runs at
about 35MB, but running pulldasher with bluebird runs at about 23-25MB.

Closes #76


sterlinghirsh:
CR :zap:

controllers/githubHooks.js
```diff
@@ -1,5 +1,5 @@
 var config     = require('../config'),
-    Promise    = require('promise'),
+    Promise    = require('bluebird'),
```
BaseInfinity: :b:lue :b:ird

BaseInfinity:
CR 👙

PR #138, opened 2018-05-12
Upgrade node-github to @octokit/rest

### LINKED ISSUE #75: Update node-github to @octokat/rest
We forked the `node-github` module a while ago when we were actively adding new webhook responders that hadn't been implemented in the upstream version.

All of our forked changes made it into upstream (yay!) so our fork is now only behind upstream (instead of ahead and behind at the same time).

I did a fast-forward merge on our repo, but we might as well just load the module from npm. If we want to fork again we can do so later.

This might be a good amount of work, we're about two years of features behind the upstream, so things are probably going to break.

copperwall:
A long time ago we forked `node-github` in order to add some changes to it. We made a pull request on the upstream and then forgot to track the upstream once our forked changes made it back in.

Our fork is about 10+ major versions behind and at least one package rename, so this adds the necessary changes to track the current version of `@octokit/rest`.

In order to cause the least number of changes to Pulldasher, this only updates the underlying SDK calls underneath our own API in `git-manager.js`. Once this is stable for a little while we can look into removing our secondary API in favor of just using `@octokit/rest` directly.

Closes #75

### DIFF
--- a/bin/refresh-open-pulls
+++ b/bin/refresh-open-pulls
@@ -5,6 +5,6 @@ var refresh = require('../lib/refresh');
 var db = require('../lib/db');
 
 refresh.openPulls()
-.done(function() {
+.then(function() {
    db.end();
 });
--- a/lib/db-manager.js
+++ b/lib/db-manager.js
@@ -113,8 +113,8 @@ var dbManager = module.exports = {
          // `pullNumber` will be null if we are updating the build status of a commit
          // which is no longer a pull's head commit.
          if (pullNumber === null) {
-            debug('updateCommitStatus: commit no longer HEAD of Pull #%s in repo %s',
-                pullNumber, updatedStatus.data.repo);
+            debug('updateCommitStatus: commit %s is not the HEAD of any pull, none to refresh in repo %s',
+               updatedStatus.data.sha, updatedStatus.data.repo);
             return;
          }
 
--- a/lib/db-manager.js
+++ b/lib/db-manager.js
@@ -104,26 +104,23 @@ var dbManager = module.exports = {
     */
    updateCommitStatus: function(updatedStatus) {
       debug('Calling `updateCommitStatus` for commit %s in repo %s',
-       updatedStatus.data.commit, updatedStatus.data.repo);
+       updatedStatus.data.sha, updatedStatus.data.repo);
       var dbStatus = new DBStatus(updatedStatus);
-      var number = this.getPullNumberFromSha(dbStatus.data.repo, dbStatus.data.commit);
 
       return dbStatus.save().
-      then(function() {
-         return number;
-      }).
-      then(function(n) {
-         // `n` will be null if we are updating the build status of a commit
+      then(() => this.getPullNumberFromSha(updatedStatus.data.repo, updatedStatus.data.sha)).
+      then(function(pullNumber) {
+         // `pullNumber` will be null if we are updating the build status of a commit
          // which is no longer a pull's head commit.
-         if (n === null) {
+         if (pullNumber === null) {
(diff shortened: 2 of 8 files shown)


lib/db-manager.js
```diff
@@ -104,26 +104,23 @@ var dbManager = module.exports = {
...
-         // `n` will be null if we are updating the build status of a commit
+      then(() => this.getPullNumberFromSha(updatedStatus.data.repo, updatedStatus.data.sha)).
+      then(function(pullNumber) {
+         // `pullNumber` will be null if we are updating the build status of a commit
          // which is no longer a pull's head commit.
-         if (n === null) {
+         if (pullNumber === null) {
             debug('updateCommitStatus: commit no longer HEAD of Pull #%s in repo %s',
-                number, updatedStatus.data.repo);
+                pullNumber, updatedStatus.data.repo);
```
danielbeardsley: This was broken before, but the message still doesn't make a whole lot of sense (`pullNumber` is always null here). Maybe just `commit is not the HEAD of any pull, none to refresh`

bin/refresh-open-pulls
```diff
@@ -5,6 +5,6 @@ var refresh = require('../lib/refresh');
 var db = require('../lib/db');
 
 refresh.openPulls()
-.done(function() {
+.then(function() {
```
danielbeardsley: Pretty sure this was `done` so that unhandled rejected promises still get reported instead of disappearing. Thoughts?

danielbeardsley:
Thank you so much!!

CR 👍 

but deploy_block 👍  on responding to some comments.

lib/git-manager.js
```diff
@@ -373,66 +342,43 @@ function parseRepo(repo) {
...
-      }, repo)).then(function(commits) {
-         if (commits.length === 1) {
-            resolve(commits[0]);
-         } else {
-            return reject();
-         }
-      });
-   });
+   return github.repos.getCommit(params({ sha }, repo))
+                      .then(res => res.data);
```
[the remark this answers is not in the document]
copperwall: We only use `getCommit` in `module.exports.parse` and we don't have any promise catch handling on the response from this function, so I don't think it really has an effect. The shape of the response is now a single object instead of a list too, so the error case doesn't seem relevant anymore.

copperwall:
un_deploy_block ⭐️

danielbeardsley:
CR 👍

addison-grant:
QA :sunny:

1. Cloned the repo locally
2. Copied over the config from @copperwall 
3. `docker-compose up`ped
4. Saw it running at `localhost:8080`!

At this point pulldasher was empty

5. Ran `docker exec -it pulldasher_web_1 /bin/bash` to log into the instance
6. Ran `bin/refresh-open-pulls`
7. Logged out and ran `docker restart pulldasher_web_1`
8. After it was done restarting, I opened the URL again, and saw it load
9. It looked nearly identical to pulldasher.cominor.com 
   * A little different because a few pulls had opened since I refreshed the pulls

Also, @copperwall mentioned the branch has been running in dev for a while today.

lib/git-manager.js
```diff
@@ -228,26 +209,16 @@ module.exports = {
...
-   debug("Got %s results", currentPage.length);
-   return new Promise(function (resolve, reject) {
-      allResults = allResults || [];
-      allResults.push(currentPage);
-
-      if (!github.hasNextPage(currentPage)) {
-         debug("No more results");
-         return resolve(_.flatten(allResults));
-      }
+async function paginate(res) {
```
sterlinghirsh: This feels like depaginate to me

sterlinghirsh:
CR :zap:

PR #148, opened 2019-04-12
Issue Refreshing: fix it

danielbeardsley:
There's actually a bunch of fixes here all detailed in the commit
messages:

* Use bluebird promises in one last place
* Use correct octokit api for fetching issues
* Re-add the accidentally rate limit
* Add logging by default to the Dockerfile
* Template rendering: turn off useless debug mode

### DIFF
--- a/controllers/main.js
+++ b/controllers/main.js
@@ -8,7 +8,7 @@ module.exports = {
          socketToken: socketAuthenticator.getTokenForUser(req.user),
          socketVersion: socketio,
          user: req.user.username,
-         debugTools: true,
+         debugTools: config.debug,
       });
    }
 };
--- a/controllers/main.js
+++ b/controllers/main.js
@@ -8,7 +8,7 @@ module.exports = {
          socketToken: socketAuthenticator.getTokenForUser(req.user),
          socketVersion: socketio,
          user: req.user.username,
-         debug: config.debug
+         debugTools: true,
       });
    }
 };
--- a/lib/git-manager.js
+++ b/lib/git-manager.js
@@ -31,7 +31,7 @@ module.exports = {
     */
    getPull: function(repo, number) {
       debug("Getting pull %s", number);
-      return github.pullRequests.get(params({ number }, repo))
+      return rateLimit(github.pullRequests.get)(params({ number }, repo))
                                 .then(res => res.data);
    },
 
@@ -42,7 +42,7 @@ module.exports = {
     */
    getOpenPulls: function(repo) {
       debug("Getting open pulls in repo %s", repo);
-      return github.pullRequests.getAll(params({}, repo)).then(paginate);
+      return rateLimit(github.pullRequests.getAll)(params({}, repo)).then(paginate);
    },
 
    /**
@@ -52,7 +52,7 @@ module.exports = {
     */
    getAllPulls: function(repo) {
       debug("Getting all pulls in repo %s", repo);
-      return github.pullRequests.getAll(params({ state: 'all' }, repo))
+      return rateLimit(github.pullRequests.getAll)(params({ state: 'all' }, repo))
                                 .then(paginate);
    },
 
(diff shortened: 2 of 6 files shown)


lib/git-manager.js
```diff
@@ -77,7 +77,7 @@ module.exports = {
    getOpenIssues: function(repo) {
       const searchParams = params({}, repo);
       debug("Getting open issues");
-      return github.issues.getAll(searchParams)
+      return github.issues.getForRepo(searchParams)
```
andyg0808: We might actually want to use [`listForRepo`](https://octokit.github.io/rest.js/#api-Issues-listForRepo), because it doesn't seem like `getForRepo` is documented for issues in those API docs.

lib/git-manager.js
```diff
@@ -91,7 +91,7 @@ module.exports = {
    getAllIssues: function(repo) {
       const searchParams = params({state: 'all'}, repo);
       debug("Getting all issues");
-      return github.issues.getAll(searchParams)
+      return github.issues.getForRepo(searchParams)
```
andyg0808: Same as above.

lib/git-manager.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
     */
    getPull: function(repo, number) {
       debug("Getting pull %s", number);
-      return github.pullRequests.get(params({ number }, repo))
+      return rateLimit(github.pullRequests.get)(params({ number }, repo))
```
andyg0808: There's a little piece of me wonders if we should iterate over the namespaces in `github` and the functions in each namespace, and apply the `rateLimit` to each of them. I suspect it might be more complex than it's worth, thought. What do you think?

controllers/main.js
```diff
@@ -8,7 +8,6 @@ module.exports = {
          socketToken: socketAuthenticator.getTokenForUser(req.user),
          socketVersion: socketio,
          user: req.user.username,
-         debug: config.debug
```
andyg0808: Nice find!

andyg0808:
Thanks! This looks really good!

lib/git-manager.js
danielbeardsley: This is the right name to use for v15.x: https://github.com/octokit/rest.js/blob/15.x/lib/routes.json#L3768-L3832

Looks like they changed it in v16 and moved everything to its own file / repo: https://github.com/octokit/routes/blob/master/routes/api.github.com/issues/list-for-repo.json

lib/git-manager.js
danielbeardsley: Thought about it, but I don't see us using very many more over time and it seems more likely that we'd mess that up, than just to forget one here.

lib/git-manager.js
andyg0808: Do I misunderstand? It seems like [this](https://github.com/octokit/rest.js/blob/15.x/lib/routes.json#L3379-L3382) is saying `getForRepo` is deprecated in v15.x?

lib/git-manager.js
andyg0808: Fair.

controllers/main.js
danielbeardsley: Turns out I broke airplane mode though. Fixed!

controllers/main.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
...
+         debugTools: true,
```
andyg0808: What about
```suggestion
         debugTools: config.debug,
```
so it can still be configured?

andyg0808:
CR 👍 
A couple comments, one about deprecation of that `getFromRepo` function and the other about leaving `debugTools` configurable.

lib/git-manager.js
danielbeardsley: Wow, you're right! But the version in our package.lock.json is 15.4.1 which *didn't* have that. I made this change by looking at the installed version in `node_modules`. We could bump the version requirement, but I'd rather not right now.

andyg0808: Works for me. Thanks for checking!

andyg0808:
CR 👍

sterlinghirsh:
CR :zap:

PR #149, opened 2019-09-14
Socket.io reconnection: stop refreshing the page

danielbeardsley:
When you focus away from the tab for a while and then come back,
the page refreshes... Wat?

Turns out chrome kills the connection (or something). After coming back,
socket.io fires a 'connect' event. We then fire the 'auth' event
to the backend thinking this is the first time we connected.
The server can't make sense of our nonce and so returns a "fail,
please refresh" response.

Now, let's just only ever listen for one auth event per server-side
connection. So long as the server thinks they're still connected,
we don't care if the client thinks it's connecting for the first time
or reconnecting after being away.

Also, add a manual 'build' npm command. Previously, it only happened on
post-commit.

### DIFF
--- a/public/js/socket.js
+++ b/public/js/socket.js
@@ -13,5 +13,4 @@ socket.on('connect', function() {
    socket.emit('authenticate', token);
 });
 
-
 export default socket;
--- a/public/js/socket.js
+++ b/public/js/socket.js
@@ -13,4 +13,5 @@ socket.on('connect', function() {
    socket.emit('authenticate', token);
 });
 
+
 export default socket;
--- a/app.js
+++ b/app.js
@@ -77,7 +77,7 @@ io.sockets.on('connection', function (socket) {
       socket.disconnect();
    }, unauthenticated_timeout);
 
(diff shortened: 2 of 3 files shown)


public/js/socket.js
```diff
@@ -13,4 +13,5 @@ socket.on('connect', function() {
    socket.emit('authenticate', token);
 });
 
+
```
jarstelfox: Not needed

jarstelfox:
CR :cactus:

sctice:
It took me a while to understand the problem and solution, but I think I'm on board now. I think I'd prefer that the client know the difference between an initial connection and a reconnect, and that it not try to authenticate with the old nonce on a reconnect, but I trust that that's difficult or more complex, and that this works fine. CR 🎏

jarstelfox:
CR 🌵

danielbeardsley:
QA 👍 

un_deploy_block 👍 

I merged this and the other pull together and ran the app. It served the main page will all the data just fine. I redelivered some webhooks and it did the right thing.

PR #151, opened 2019-09-26
Authentication: Don't check org membership if not required

copperwall:
Hey everyone 👋, I was playing around with Pulldasher and ran into an issue with GitHub authentication where it would still check my org membership even if I left the requireOrg config variable as false.

This makes a change to not check a user's org membership if the
requireOrg field in config.js is falsy. Currently Pulldasher always
makes a call to GitHub to check the user's membership, when it shouldn't
if the user doesn't want to require an org.

I tested this with a requireOrg value of `iFixit`, in which I was not allowed to view my local pulldasher website, and with a value of `false`, in which I could view my local instance of pulldasher.

### DIFF
--- a/lib/authentication.js
+++ b/lib/authentication.js
@@ -33,7 +33,7 @@ module.exports = {
       app.get('/auth/github', passport.authenticate('github'));
       app.get('/auth/github/callback', function(req, res, next) {
          passport.authenticate('github',
-         function(err, user, info) {
+         function(err, user) {
             if (err) {
                debug("Error: " + err);
                return next(err);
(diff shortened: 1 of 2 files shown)


danielbeardsley:
Hi Chris!!

CR 👍

lib/authentication.js
```diff
@@ -45,19 +45,30 @@ module.exports = {
...
+                  debug("Authenticated with org: " + user.username);
+                  req.logIn(user, function() {
+                     debug('Got login from ' + user.username);
+
+                     // Redirect user to original URL. Fallback to '/' if DNE.
+                     const redirectUrl = req.session.auth_redirect;
+                     delete req.session.auth_redirect;
+
+                     res.redirect(redirectUrl || '/');
+                  });
```
andyg0808: I wish this didn't have to be duplicated with the code below, but eh.

If you really want to, you could pull this bit out and give it a name, then reuse it in both places. Unless something cleverer occurs to you.

andyg0808:
CR :+1: 

Hi and thanks!

lib/authentication.js
copperwall: Yeah I totally see that. My first attempt was to do something like

```js
let afterMembershipValidation = Promise.resolve();

if (config.github.requireOrg) {
   afterMembershipValidation = confirmOrgMembership(user);
}

afterMembershipValidation.then(() => {
   // The shared log in and redirect logic.
});
```

The only thing was the function passed to the `then` handler had a debug statement saying that the user was authenticated with the org, which only applies to one of those cases.

Maybe something like this?

```js
let afterMembershipValidation = Promise.resolve();

if (config.github.requireOrg) {
   afterMembershipValidation = confirmOrgMembership(user).then(
      () => debug('Authenticated with org: ' + user.username)
   );
}

afterMembshipValidation.then(() => {
   // The shared log in and redirect logic.
});
```

If having a no-op Promise is kind of hacky I'd be interested in making this some named function as well.

andyg0808: That looks fine to me. It seems kinda equivalent to assigning 0 to a variable that you might add a number to later. If it bugs you, what about putting the `Promise.resolve()` assignment in an `else` case?

addison-grant:
Looks good from my point of view.

QA :sunny:

PR #152, opened 2019-09-28
fix(authentication): deduplicate req.logIn function

copperwall:
This includes some CR changes that didn't make it into the previous PR haha.

This also adds a catch handler for the org membership check promise and
passes the error to `next` to be handled. Currently this goes to the
default error handler, but before this the unhandled promise rejection
would trigger a console warning but the response wouldn't end because the
middleware chain wouldn't continue.

Now the error message is printed out and the response is ended
successfully.

A few thoughtful CRs is enough. Worst case is login breaks and we revert.
qa_req 0

### DIFF
--- a/lib/authentication.js
+++ b/lib/authentication.js
@@ -45,9 +45,19 @@ module.exports = {
             }
 
             debug("Authenticated with github: " + user.username);
+            let afterMembershipValidation;
+
+            // If the config requires org membership, check that the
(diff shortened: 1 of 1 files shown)


andyg0808:
CR :+1:

danielbeardsley:
CR 👍

addison-grant:
We have a lot of stuff going on in QA right now, so we're `external_block`ing this for a few days, likely.

PR #161, opened 2019-11-07
Tooltips: don't santize html

### LINKED ISSUE #160: New empty tooltips
[`tooltips--allow-html`](https://github.com/iFixit/pulldasher/compare/tooltips--allow-html)

We have little tooltips for the CR/QA checkoff icons like `No signoffs yet!` and `No QA required!`.

Now the tooltips are showing up even when there's no content for them.

I am guessing a dependency from the latest dependabot pulls changed this.

danielbeardsley:
The default santization has a whitelist of html element types.
Some of the ones we use (like \<table\>) aren't on the list.

Let's drop sanitization all together. We're constructing safe
html right now (I checked, we're only using node.text() not
node.html()), so let's consider that good enough.

Also, run 'npm update' and commit the result.
Let's stay ontop of those versions!

CC @addison-grant

Closes #160


jarstelfox:
CR 🉑

cdcline:
CR 👍

addison-grant:
Oops, I didn't realize this before, but the tooltips are supposed to also show the names of the reviewers!

I'll start up a pulldasher from this branch to test.

addison-grant:
Tooltips appear fixed!! No obvious regressions that I can see.

QA :sunny:

PR #162, opened 2019-11-14
Multiple Commit Statuses: Implement MVP

### LINKED ISSUE #72: CI Builds: Support multiple

danielbeardsley:
Support multiple statuses per commit:

* Injest them from the webhooks
* Save them to the DB
* Load them from the DB
* Request them from the API
* Maintain them in the memory model
* Update the UI
  * Split the existing green/red CI status bar into vertical pills
    with one per status

image

Closes #72

### DIFF
--- a/lib/git-manager.js
+++ b/lib/git-manager.js
@@ -365,8 +365,9 @@ function getCommit(repo, sha) {
 
 function getCommitStatuses(repo, ref) {
    debug("Getting commit status for %s", ref);
-   return rateLimit(github.repos.getStatuses)(params({ ref }, repo))
-                      .then(res => res.data)
+   return rateLimit(github.repos.getCombinedStatusForRef)(params({ ref }, repo))
+      .then(res => res.data.statuses)
+      .then(statuses => statuses || [])
 }
 
 /**
--- a/lib/git-manager.js
+++ b/lib/git-manager.js
@@ -112,7 +112,7 @@ module.exports = {
       var reviewComments = getPullReviewComments(repo, githubPull.number);
       var comments = getIssueComments(repo, githubPull.number);
       var headCommit = getCommit(repo, githubPull.head.sha);
-      var commitStatus = getCommitStatus(repo, githubPull.head.sha);
+      var commitStatuses = getCommitStatuses(repo, githubPull.head.sha);
       var events = getIssueEvents(repo, githubPull.number);
       // Only so we have the canonical list of labels.
       var ghIssue = module.exports.getIssue(repo, githubPull.number);
@@ -122,14 +122,14 @@ module.exports = {
       return Promise.all([reviewComments,
                           comments,
                           headCommit,
-                          commitStatus,
+                          commitStatuses,
                           events,
                           ghIssue])
        .then(function(results) {
          var reviewComments = results[0],
              comments       = results[1],
              headCommit     = results[2],
-             commitStatus   = results[3],
+             commitStatuses = results[3],
              events         = results[4],
              ghIssue        = results[5];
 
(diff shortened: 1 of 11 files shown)


models/pull.js
```diff
@@ -7,11 +7,11 @@ var debug = require('../lib/debug')('pulldasher:pull');
 var DBPull = require('./db_pull');
 var getLogin = require('../lib/get-user-login');
 
-function Pull(data, signatures, comments, commitStatus, labels) {
+function Pull(data, signatures, comments, commitStatuses, labels) {
    this.data = data;
    this.signatures = signatures || [];
    this.comments = comments || [];
-   this.commitStatus = commitStatus;
+   this.commitStatuses = commitStatuses || [];
```
davidrans: Isn't this always an `array` now?

models/pull.js
```diff
@@ -114,7 +114,7 @@ Pull.prototype.isOpen = function() {
  *    'deploy_block'  : An array containing the last 'deploy_block' signature if pull is deploy blocked,
  *                       or an empty array
  *    'ready'         : A boolean indicating whether the pull is ready to be deployed.
- *    'commit_status' : A Status object or null.
+ *    'commit_statuses' : A Status object or null.
```
davidrans:
```suggestion
 *    'commit_statuses' : An array of Status objects
```

davidrans:
CR ☕️

models/pull.js
danielbeardsley: You're right.

models/pull.js
danielbeardsley: Fixed!

models/pull.js
danielbeardsley: Fixed!

davidrans:
CR ☕️

djmetzle:
CR :vertical_traffic_light:

mlahargou:
@addison-grant and I attempted to QA this.

1) Run `docker rm -f pulldasher`
1) Open `specific/dev/startup-scripts/install-pulldasher` in server-templates
2) Change the `VERSION` to `origin/commit-statues--suport-more-than-one`
3) Add `read -p Paused` after the copy config (line 23)
4) Run `bash specific/dev/startup-scripts/install-pulldasher`
5) When the script gets paused, open up `config.js` and change the db to point to cominor
6) Hit enter on the install script and the new pulldasher should be up and running

I followed the above commands and noticed that we are now able to have different statuses.  

I tried to edit the cominor metrics table in order to see if > 2 statuses looked decent and if the other states (error) looked decent.  When I applied the changes to the DB, they were not reflected on pulldasher.cominor.com.

I attempted `docker restart pulldasher`, and `bin/refresh-all-pulls` in order to try to get pulldasher to refresh.  Neither of these worked.  

I can say that I was able to see what Danny saw: Two statuses with the correct description, context, state, and url for a single pull/commit.  QA 📱 on that but deploy_block 📱 because I **would** like to test a few things further like:

1) If a commit has two success statuses, does it go to the correct columns?
2) If a commit has 3 statuses, does the UI look decent?
3) If a commit has all combinations of states, does it behave correctly?

danielbeardsley:
Thanks @addison-grant and @mlahargou !!

Regarding the remaining QA bullets, I wouldn't worry about those ones, their blast radius is low, 
 they are easy to fix after the fact, and they are easy for CR to verify that it's correct (`build_succeeded()` and friends are easy to CR).

What I *would* like more QA on (and is harder to validate / prove that it's done right) is if incoming webhooks produce the desired state change in the DB / UI. This will require running pulldasher with `DEBUG='*'` (also against the devdb as before) and then getting github to fire some notifications. You can go to github's GUI and re-send any past notifications.

danielbeardsley:
dev_block 👍   found a bug. Using the wrong github API endpoint to retrieve statuses.

lib/git-manager.js
```diff
@@ -365,8 +365,9 @@ function getCommit(repo, sha) {
 
 function getCommitStatuses(repo, ref) {
    debug("Getting commit status for %s", ref);
-   return rateLimit(github.repos.getStatuses)(params({ ref }, repo))
-                      .then(res => res.data)
+   return rateLimit(github.repos.getCombinedStatusForRef)(params({ ref }, repo))
```
djmetzle: Neat!

djmetzle: Docs are here for the interested:
Ref: https://octokit.github.io/rest.js/

djmetzle:
More CR :traffic_light:

danielbeardsley:
un_dev_block 👍 

QA 👍

davidrans:
CR ☕️

danielbeardsley:
un_deploy_block 👍

PR #165, opened 2019-11-21
Allow overriding the config path from the ENV

djmetzle:
In order to mount the pulldasher configuration as a secret, we need to be able to mount a volume, not a single file. The `config.js` file was hardcoded here to `/opt/pulldasher/config.js`. Let's allow optionally specifying the path on the environment.

cr_req 1
qa_req 0

CC @danielbeardsley


jarstelfox:
CR 🌵

PR #166, opened 2020-03-04
Pull Diff: store data in DB

davidrans:
This will give us the ability to track `additions`, `deletions`, and
`changed_files` on pulls.

QA
----
I ran a local instance of pulldasher and confirmed that these fields
were being filled in in the `pulls` table for new pulls. Anything else
we should check?

### DIFF
--- a/migrations/0012-add-pull-diff-info.sql
+++ b/migrations/0012-add-pull-diff-info.sql
@@ -1,4 +1,5 @@
 ALTER TABLE `pulls`
+ADD COLUMN `commits` int(11) unsigned DEFAULT NULL,
 ADD COLUMN `additions` int(11) unsigned DEFAULT NULL,
 ADD COLUMN `deletions` int(11) unsigned DEFAULT NULL,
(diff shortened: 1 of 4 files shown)


djmetzle:
@davidrans 
> Anything else we should check?

Ref: https://developer.github.com/v3/pulls/#get-a-single-pull-request

Can we get `commits` and `comments` too?

andyg0808:
CR :+1:

davidrans:
> Can we get `commits` and `comments` too?

Good idea @djmetzle. I added `commits`, but we already have comments in another table.

jarstelfox:
CR 🌵 I wasn't aware we tracked these data points... time to start querying 😝

djmetzle:
CR :gear:

k0rvusk0r4x:
> Anything else we should check?

Have we tested/do we have an idea of what we expect if a change is made while something in the chain is down? Will it get picked up and soon as things are working again? Do we have a script we can run to backfill?

What happens if a "change" is made that actually changes nothing?

PR #168, opened 2020-04-03
Commit statuses: allow being explicit

### LINKED ISSUE #167: Commit Statuses: Allow configuring which statuses are expected / required
[`ci-status--required`](https://github.com/iFixit/pulldasher/compare/ci-status--required)

We'd like to be able to explicitly list which commit status are required before we mark a pull as green. This list should be per-repo.

danielbeardsley:
Allow configuring the required commit statuses per repo so we don't
accidentally show something as green when it's missing some CI statuses.

This is backward compatible, existing configs will be interpreted in the
same way. Implemented this by sending the list of repos with their
requirements on the initial socker.io connection.

The config docs say it well:

    // Just listing the repo as a string indicates pulldasher should consider
    // CI passing if all the statuses that exist are passing and that there
    // is at least one.
    "owner/repo",
    // Listing them like this allows being more explicit with the commit
    // statuses
    {name: "owner/otherRepo", requiredStatuses: ["tests", "build", "codeClimate"]}

Closes #167


lib/authentication.js
```diff
@@ -76,6 +80,10 @@ module.exports = {
       // Ensure index is authenticated.
       app.get('/', function (req, res, next) {
          if (req.isAuthenticated()) { return next(); }
+         if (process.env.MOCK_AUTH_AS_USER) {
+            req.user = {username: process.env.MOCK_AUTH_AS_USER};
```
sterlinghirsh, line 84:
```suggestion
         if (FAKE_USER) {
            req.user = {username: FAKE_USER};
```

lib/config.js
```diff
@@ -0,0 +1,3 @@
+var configPath = process.env.CONFIG_PATH || '../config';
```
sterlinghirsh, line 1: It's a little confusing to have 2 config.js files. lib/config.js requires ../config.js. Anyone loading the config file directly may accidentally miss this file, and it's a hard mistake to spot.

sterlinghirsh:
CR :zap: dev_block :zap: on a few small comments. Feel free to not rename config.js if you don't think it's worth.

lib/config.js
danielbeardsley: Good point. I'll rename this one.

danielbeardsley:
QA 👍 

Ran this in development and tested several different iterations of `requiredStatuses` and they all worked.

danielbeardsley:
un_dev_block 👍

lib/authentication.js
danielbeardsley: Fixed!

sterlinghirsh:
CR :zap:

danielbeardsley:
Re QA 👍

cdcline:
CR 👍

jarstelfox:
Once this is deployed what is the process to upgrade pulldasher?

addison-grant:
> Once this is deployed what is the process to upgrade pulldasher?

Steps 3 to 5 in https://overlord.dozuki.com/c/Pulldasher#Section_Adding_a_repo

PR #171, opened 2020-04-13
config.repos: normalize repo list on load

danielbeardsley:
In a previous pull, config.repos learned to handle an array of objects
instead of just an array of strings. Not all usages of the config.repos
setting were updated to handle the different ways of representing the
data.

Instead of handling both formats in each usage, let's normalize when
loading the config.

Connect #167


sterlinghirsh:
CR :zap:

djmetzle:
CR :zap:

addison-grant:
I think just booting pulldasher should test this, right? Maybe we should also test booting the app with a newer style repo specification (object instead of string).

addison-grant:
✅ 

I think it's good! QA :sunny:

PR #172, opened 2020-04-13
Pulls: refresh open pulls on app start

danielbeardsley:
Previously we hadn't done this because a boot-loop could cause you
to burn through your API rate limit pretty quickly. But in practice
that doesn't happen anymore, and this is worth it.

### DIFF
--- a/lib/config-loader.js
+++ b/lib/config-loader.js
@@ -1,3 +1,13 @@
 var configPath = process.env.CONFIG_PATH || '../config';
-module.exports = require(configPath);
+const config = require(configPath);
+module.exports  = config;
 
+config.repos = config.repos.map(normalizeRepo);
+
+function normalizeRepo(repo) {
+   if (typeof repo == 'string') {
+      return {name: repo};
+   } else {
+      return repo;
+   }
+}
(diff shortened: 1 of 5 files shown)


lib/config-loader.js
```diff
@@ -1,3 +1,13 @@
 var configPath = process.env.CONFIG_PATH || '../config';
-module.exports = require(configPath);
+const config = require(configPath);
+module.exports  = config;
 
+config.repos = config.repos.map(normalizeRepo);
```
djmetzle, line 5: This seems verbose for avoiding ES6 sake?
```suggestion
config.repos = config.repos.map(repo => {
   (typeof repo == 'string') ? {name: repo} : repo;
 });
```

djmetzle:
CR :arrows_clockwise:

davidrans:
CR :coffee:

davidrans:
Looks like this ought to be deploy_block :hand:  on https://github.com/iFixit/pulldasher/pull/171

addison-grant:
Things booted fine. I am guessing that's practically good enough? QA :sunny:

PR #177, opened 2020-06-17
Support more emojis in signoffs

davidrans:
QA
---
Copy this new regex to from the example config to the real config, and
make sure it:
 - works with emojis that our current regex doesn't like 🪂
 - doesn't match things that aren't emojis

cr_req 1

CC @ianrohde

On Deploy
---
Copy this new regex to from the example config to the real config.

Closes iFixit/pulldasher#176

Issue #174, opened 2020-05-06
Pulldasher:master is currently broken on tooltip dependency


addison-grant:
At least, that's what it seems like. Attempting to deploy master after merging in #173 caused Pulldasher to crash on one thing, and trying to fix that revealed another problem.

First attempt at fix: 2d8af13f5a1b1bde28cc9c695ecfe1c5c66228fe
The second problem that arose was a browser-side error on trying to render tooltips. This manifested in the page failing to load completely. There were some JS errors revealed in the browser console.

This is the only stacktrace I have captured related to the tooltips:

```
Navigated to https://pulldasher.cominor.com/
VM5265 bundle.js:41 Uncaught TypeError: r.hasOwnProperty is not a function
    at u.getOptions (VM5265 bundle.js:41)
    at u.init (VM5265 bundle.js:41)
    at new u (VM5265 bundle.js:41)
    at HTMLDivElement.<anonymous> (VM5265 bundle.js:41)
    at Function.each (VM5265 bundle.js:14)
    at k.fn.init.each (VM5265 bundle.js:14)
    at k.fn.init.t.fn.tooltip (VM5265 bundle.js:41)
    at Object.addTooltip (VM5265 bundle.js:25)
    at VM5265 bundle.js:41
    at Function.y.each.y.forEach (VM5265 bundle.js:25)
```

This is another I saw at one point:

![image](https://user-images.githubusercontent.com/10889879/81239041-88cb4a00-8fb8-11ea-8cd4-cd84b3d2e124.png)

CC @danielbeardsley

addison-grant:
When we figure this out, unfreeze https://github.com/iFixit/pulldasher/pull/181.

ardelato:
@addison-grant where do we stand on this? It has been a while since this issue was created but now it seems that this is impeding #200

addison-grant:
> where do we stand on this?

Not sure. It hasn't been worked on as of now, as far as I know.

addison-grant:
Is there something incorrect about this GH issue? I am seeing `master` deployed currently, I think.
```
agrant@cominor.com ~/pulldasher (master)
$ git show
commit 5b13321e96072aa54a85767817ab22c9831e9038 (HEAD -> master, origin/master, origin/HEAD)
Merge: edfd343 f702edd
Author: Addison Grant <addison-grant@users.noreply.github.com>
Date:   Wed Dec 23 10:25:11 2020 -0800

[...]

PR #180, opened 2020-06-23
Temporarily revert jQuery bump

addison-grant:
Background
===

@davidrans tried deploying his recent change #177 but we forgot about an unresolved rollback that happened a month or two ago.

Let's look into why the jQuery bump caused issues, but let's run the
other recent changes in the mean time.

This should allow us to use the most recent changes aside from those in https://github.com/iFixit/pulldasher/pull/173

QA
===

Run this branch and see if you can spot any immediately obvious problems. If there's nothing to glaring, let's merge this and let it run in the wild.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5108,9 +5108,9 @@
       "optional": true
     },
     "jquery": {
-      "version": "3.5.0",
-      "resolved": "https://registry.npmjs.org/jquery/-/jquery-3.5.0.tgz",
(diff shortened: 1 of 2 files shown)


sctice:
CR 🎏

addison-grant:
Issue written when we first ran into the jQuery problem. There might be more useful info in @danielbeardsley's noggin. I'll try to update it with better details soon. https://github.com/iFixit/pulldasher/issues/174

jarstelfox:
CR :cactus:

batbattur:
testing 
dev_block 🪂 this is testing

batbattur:
un_dev_block 🪂 test

batbattur:
QA 👍

PR #183, opened 2020-06-25
Allow one signature per person

andyg0808:
Let's restrict the number of signatures each person can make to one. This will be slightly more bother when carrying over a CR, but it will reduce the risk of accidentally deploying a pull which only one person has signed off on.

We encountered an incident where a person signed off twice, once before a commit was pushed and once after, but because the commit date was before the first signoff date, both signoffs were considered valid. The problem with a signoff not being invalidated by a commit which is pushed after the signoff is still not solved, but this change at least reduces the risk of such an incident resulting in insufficiently-reviewed code being deployed.

### DIFF
--- a/public/js/Pull.js
+++ b/public/js/Pull.js
@@ -27,18 +27,13 @@ _.extend(Pull.prototype, {
             if (!users[signature.data.user.id]) {
                if (signature.data.active) {
                   groups.current.push(signature);
-                  users[signature.data.user.id] = true;
-
-                  if (utils.mySig(signature)) {
-                     groups.user = signature;
-                  }
                } else {
                   groups.old.push(signature);
-                  users[signature.data.user.id] = true;
+               }
+               users[signature.data.user.id] = true;
 
-                  if (utils.mySig(signature)) {
-                     groups.user = signature;
-                  }
+               if (utils.mySig(signature)) {
+                  groups.user = signature;
                }
             }
          });
--- a/public/js/Pull.js
+++ b/public/js/Pull.js
@@ -24,19 +24,21 @@ _.extend(Pull.prototype, {
          groups.user = null;
 
          signatures.forEach(function(signature) {
-            if (signature.data.active) {
-               groups.current.push(signature);
-               users[signature.data.user.id] = true;
+            if (!users[signature.data.user.id]) {
+               if (signature.data.active) {
+                  groups.current.push(signature);
+                  users[signature.data.user.id] = true;
 
-               if (utils.mySig(signature)) {
-                  groups.user = signature;
-               }
-            } else if (!users[signature.data.user.id]) {
-               groups.old.push(signature);
-               users[signature.data.user.id] = true;
+                  if (utils.mySig(signature)) {
+                     groups.user = signature;
+                  }
+               } else {
+                  groups.old.push(signature);
+                  users[signature.data.user.id] = true;
 
-               if (utils.mySig(signature)) {
-                  groups.user = signature;
+                  if (utils.mySig(signature)) {
+                     groups.user = signature;
+                  }
                }
             }
          });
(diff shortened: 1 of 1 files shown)


public/js/Pull.js
```diff
@@ -24,15 +24,12 @@ _.extend(Pull.prototype, {
          groups.user = null;
 
          signatures.forEach(function(signature) {
-            if (signature.data.active) {
-               groups.current.push(signature);
-               users[signature.data.user.id] = true;
-
-               if (utils.mySig(signature)) {
-                  groups.user = signature;
+            if (!users[signature.data.user.id]) {
```
jarstelfox, line 27: I would invert this and early return to reduce nesting.

public/js/Pull.js
```diff
@@ -24,15 +24,12 @@ _.extend(Pull.prototype, {
...
+               if (signature.data.active) {
+                  groups.current.push(signature);
+               } else {
+                  groups.old.push(signature);
```
jarstelfox, line 31: Reduce nesting:
```suggestion
               const groups = signature.data.active ? groups.current : groups.old;
               groups.push(signature);
```

jarstelfox, line 31: Reduce nesting:
```suggestion
               const groups = signature.data.active ? groups.current : groups.old;
               groups.push(signature);
```

FYI: I can not include the last brace in my suggestion so do this by hand please

djmetzle:
CR :gear: though @jarstelfox has some good cleanup ideas.

public/js/Pull.js
davidrans: This is fewer lines, but not more readable IMO :shrug:

jarstelfox: IMO it's more readable. Namely, you don't have to keep the state in your head. Every time I see an if, I mentally note that we are forking the state and need to consider other cases. Between inversion and ternaries, it becomes more clear to me that we are handling + edge case and picking a bucket.

That being said it's up to Andrew here

sctice:
I'm finding this logic a bit difficult to follow, so I may very well be mistaken, but it looks to me like the new version could fail to recognize an active signature for someone who also has an inactive signature that was processed first. It looks to me like whichever case we encounter first, inactive or active, that's the end of the story for that user. For the same reasons, I think we're changing how the `groups.user` signature gets assigned.

I had a hard time when just looking at the diff, so in case it helps anyone else, here's the previous version in full:

https://github.com/iFixit/pulldasher/blob/236d7bf246be021570268ea83e38967637071a38/public/js/Pull.js#L14-L45

CR 🎏 
dev_block ✋ on what looks to me like an unintended change in behavior.

andyg0808:
dev_block :+1: 
on comments. Good points!

andyg0808:
@sctice I believe the signatures are handled ordered by date descending. I have two reasons for this belief:
1. If they were not, the existing code would potentially add an "expired" sign off and then a "current" signoff for the same user, if they were encountered in that order.
2. The code to fetch signatures from the DB sorts them in descending date order:
https://github.com/iFixit/pulldasher/blob/e040c3b2b741d3af11b3ec7d8cfe91a5c4a3036a/lib/db-manager.js#L345-L356
Although I can't guarantee that that's how the data is fetched on every code path, I seem to remember that the DB is integral to how the backend works, so I'm guessing that method may, in fact, be the only way signature data gets to the frontend.

As far as I've seen, the pulls are refreshed as objects, so I don't think there's any update of the signature list item-by-item; I think we're sending the entire new pull to the frontend every time a signature is updated.

public/js/Pull.js
andyg0808: Good idea!
3a87c82

public/js/Pull.js
andyg0808: I'm fine with this the way it is.

andyg0808:
un_dev_block :+1:

djmetzle:
CR :pen:

jyee27:
CR :coffee:

PR #186, opened 2020-07-15
Pulldasher: Clarify webhook `secret` param

andyg0808:
The docs on this weren't clear on all the requirements to get the webhook working.


zdmitchell:
CR :guitar:

djmetzle:
CR :arrow_right_hook:

danielbeardsley:
Thakns!

Soonish we should just use Github's method of validating the request is coming from them:
https://developer.github.com/webhooks/securing/#validating-payloads-from-github

Same idea (a pre-shared secret) but it uses hmac on the message body instead of comparing the value directly.

batbattur:
QA 👍

PR #190, opened 2020-09-03
Github Pull hooks: refresh on more events

danielbeardsley:
For adding or removing a milestone to a pull, GH sends an "issue event"
cause every pull is *also* an issue.

Previously, we tried to use the body to update the issue in the DB
(which didn't exist). Thus we'd miss those events. Now any time we get
an issue event for a pull, we just refresh that whole pull. We have to
do this cause the body that comes with issue events is incomplete.

qa_req 0

### DIFF
--- a/controllers/githubHooks.js
+++ b/controllers/githubHooks.js
@@ -131,6 +131,22 @@ function handleIssueEvent(body) {
 
    var doneHandling = handleLabelEvents(body);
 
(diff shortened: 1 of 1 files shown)


andyg0808:
CR :+1: through 18be98e

jyee27:
CR :coffee:

danielbeardsley:
I'd vote for not bothering with QA.

PR #203, opened 2021-03-26
Improve docker-compose setup

addison-grant:
These changes were intended to make development easier. @ardelato and I are working on adding a Pull Label History table. The Pull Labels table currently only stores the current state of labels but not the history of when labels were added and removed. The goal is to get a better sense of how long it takes for PRs to go through a QA cycle, that is, when someone adds the QAing label and removes it after testing. This is an effort to help project management estimate how long it will take things to go through QA review.

The crux of this change is in `docker-compose.yml` and `config.example.js`, changing the DB settings to pull from `process.env`. For our internal use, https://github.com/iFixit/ifixit_scripts/pull/162 supports this change.

The `docker-compose.yml` change is just to make setting up your pulldasher setup a little more straightforward. The intention with pulling out the DB, port, and docker image settings is to make it more clear what is easy to configure. The `.env` and `.env.db` example files are to give an example of how to configure the docker-compose setup. 

I reorganized some operational and development scripts, and I tried to update the docs in a sensible way.

QAing
===

Try running your own pulldasher. Work with me to see how we can improve the documentation.


build-compose
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+docker-compose --env-file=.env build
```
danielbeardsley, line 3: Why do we need to run build through compose? And why do we need the .env?

docker-compose.yml
```diff
@@ -1,31 +1,34 @@
...
 
 services:
    web:
       build: .
-      entrypoint: '/opt/pulldasher/entrypoint.sh'
+      image: ${DOCKER_IMAGE_NAME}
+      depends_on:
+        - 'db'
+      env_file:
+        - "./.env.db"
```
danielbeardsley, line 10: Seems like a bit of trouble to separate out the db env. The default behavior is to load `.env`, so why not use that?

dev/cp-configs-to-root
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# Meant to be run from the project root
+
+cp config/.env* .
```
danielbeardsley, line 5: There is almost twice as many chars in this filename as there is in the only command it's running :-) Seems like this script is superfluous.

bin/README.md
```diff
@@ -0,0 +1,3 @@
+# directory `bin`
+
+This directory contains tools for operating pulldasher
```
danielbeardsley, line 3: Heh...  I really don't think `/bin` needs an explanation. It's effectively a linux / unix standard.

.env.example
```diff
@@ -0,0 +1,5 @@
+# Application configuration values for use with docker-compose
+DOCKER_IMAGE_NAME=<docker image name, example: pulldasher:development>
+APP_PORT=<application port number>
+DEBUG_PORT=<debug port number>
+DB_PORT=<db port on host>
```
danielbeardsley, line 5: I think these example files are a good idea, but I think it should come with defaults or a default .env file so it just works right out of the box. i.e. when running in docker-compose, don't bother opening any external ports and thus we can use the default user/password/host/database.

Same is true with the others, pick sensible defaults, but let the user override as necessary.

danielbeardsley:
Excited about this!

dev_block :+1:  on some comments!

dev/cp-configs-to-root
addison-grant: Heh. I guess this is more for my convenience while developing. Maybe I'll just keep it in my `config/` dir which is ignored by git. My one thought is, I find it useful, and maybe others would, too? If they area also keeping their test configs in the hidden `config/` dir.

addison-grant: I can also make it have a shorter name, but I wanted it to be descriptive.

bin/README.md
addison-grant: I guess so 🤷🏼‍♂️     I don't think it hurts though for those who may not be familiar with the convention.

.env.example
addison-grant: Sounds good to me, I'll make that change.

docker-compose.yml
addison-grant: Well, the thing I am calling `.env` isn't needed for the regular "production" docker run. Maybe I should call it `.env.compose`, and call this one `.env`.

build-compose
addison-grant: The `.env` file values are filled into the `docker-compose.yml` file during the build step. I tried specifying the env file in the `docker-compose.yml` but when you specify `env_file` in docker-compose, it's only used for running containers, not building, it seems.

addison-grant:
Thanks for all the feedback, @danielbeardsley! I'm going to make a couple changes, and I wonder what you think about some of my replies.

build-compose
addison-grant: Oh, and we don't need to run it through compose. It's just an option. It's a convenient wrapper on the Docker build process, and I have been using it while testing out the docker-compose setup.

bin/README.md
addison-grant: Also, the main reason for this is when you're browsing GitHub, it displays the README.md in each directory when viewing directories.

dev/cp-configs-to-root
addison-grant: Decided to remove the file. https://github.com/iFixit/pulldasher/pull/203/commits/4e08edc2ae6fed66535adfc0309c15835b65074d

addison-grant:
Getting really close. Just sorting out the environment variables and their files' naming scheme.

batbattur (commented):
Test comment

addison-grant:
I am limited on time currently. Going to close for now and we can use this issue as reference if we revisit this issue.

PR #208, opened 2021-04-30
Pulls: Add age indicator

mlahargou:
We want to be able to tell how hold a pull is at a glance. This adds the
indicator as well as adds it to the pull element.

Note: We cap the number of dots at 20 but the tooltip still shows the 
correct number of days.

image

QA:
--
I currently have it set up on https://pulldasher-dev.cominor.com.  Play
around with it! You could open a test pull to ensure it adds the proper
dots.

cc @addison-grant (We both worked on this).

### DIFF
--- a/views/standard/spec/indicators.js
+++ b/views/standard/spec/indicators.js
@@ -311,7 +311,7 @@ export default {
 function getAgeColor(days) {
    switch (Math.floor(days / 3)) {
       case 0: return 'green';
-      case 1: return '#fabd02';
+      case 1: return '#ffe62f';
       case 2: return 'orange';
       default: return 'red';
    }
--- a/views/standard/spec/indicators.js
+++ b/views/standard/spec/indicators.js
@@ -284,13 +284,13 @@ export default {
    age: function age(pull, node) {
       const timeDifference = Date.now() - new Date(pull.created_at).getTime();
       const daysSinceCreation = Math.ceil(timeDifference / (1000 * 3600 * 24));
-      const numDots = Math.min(daysSinceCreation, 20);
+      const numDots = Math.min(daysSinceCreation, 30);
       const dot = "•";
       const daysInDots = dot.repeat(numDots);
       const severityColor  = getAgeColor(daysSinceCreation);
 
       var ageContainer = $(`<div class="age_container" style="color:${severityColor}">${daysInDots}</div>`);
-      utils.addTooltip(ageContainer, `Age: ${daysSinceCreation}`);
+      utils.addTooltip(ageContainer, `Age: ${daysSinceCreation}`, 'right');
       node.append(ageContainer);
    },
 
--- a/views/standard/spec/indicators.js
+++ b/views/standard/spec/indicators.js
@@ -284,7 +284,7 @@ export default {
    age: function age(pull, node) {
       const timeDifference = Date.now() - new Date(pull.created_at).getTime();
       const daysSinceCreation = Math.ceil(timeDifference / (1000 * 3600 * 24));
-      const numDots = daysSinceCreation < 21 ? daysSinceCreation : 20;
+      const numDots = Math.min(daysSinceCreation, 20);
       const dot = "•";
       const daysInDots = dot.repeat(numDots);
       const severityColor  = getAgeColor(daysSinceCreation);
@@ -311,7 +311,7 @@ export default {
 function getAgeColor(days) {
    switch (Math.floor(days / 3)) {
       case 0: return 'green';
-      case 1: return 'yellow';
+      case 1: return '#fabd02';
       case 2: return 'orange';
       default: return 'red';
    }
--- a/views/standard/spec/indicators.js
+++ b/views/standard/spec/indicators.js
@@ -281,6 +281,18 @@ export default {
          node.append(tag);
       });
    },
+   age: function age(pull, node) {
+      const timeDifference = Date.now() - new Date(pull.created_at).getTime();
+      const daysSinceCreation = Math.ceil(timeDifference / (1000 * 3600 * 24));
+      const numDots = daysSinceCreation < 21 ? daysSinceCreation : 20;
+      const dot = "•";
+      const daysInDots = dot.repeat(numDots);
+      const severityColor  = getAgeColor(daysSinceCreation);
+
+      var ageContainer = $(`<div class="age_container" style="color:${severityColor}">${daysInDots}</div>`);
+      utils.addTooltip(ageContainer, `Age: ${daysSinceCreation}`);
+      node.append(ageContainer);
+   },
 
    // Here's an interesting duck! This indicator actually adds the event
    // handler to _the refresh button_ on each pull. If you look at the pull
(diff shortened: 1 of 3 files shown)


mlahargou:
Steps for development:
- Copied over the live pulldasher (`ls /mnt/ebs/etc/pulldasher.*`) config and .env files and put it in my cloned pulldasher repo on cominor.
- From that repo, made my changes.
- Created an image with `docker build --no-cache -t pulldasher:mike-test .`
- Created a container with `docker run -d --name=pulldasher-mike-test --publish=51515:8090 --env-file=.env.db --env MOCK_AUTH_AS_USER=fakeuser pulldasher:mike-test`
  - Note: we use 51515 because we have a backend for pulldasherdev set up in `development.vcl`

views/standard/spec/indicators.js
```diff
@@ -296,6 +308,15 @@ export default {
    }
 };
 
+function getAgeColor(days) {
+   switch (Math.floor(days / 3)) {
+      case 0: return 'green';
+      case 1: return 'yellow';
```
addison-grant, line 314:
```suggestion
      case 1: return '#fabd02';
```

`yellow` is difficult to see on the white background. This color I suggested seems better and it looks fine in dark mode, too.

GrantMatejka:
I think this is great but also agree with possible color change

CR :corn:

views/standard/spec/indicators.js
```diff
@@ -281,6 +281,18 @@ export default {
          node.append(tag);
       });
    },
+   age: function age(pull, node) {
+      const timeDifference = Date.now() - new Date(pull.created_at).getTime();
+      const daysSinceCreation = Math.ceil(timeDifference / (1000 * 3600 * 24));
+      const numDots = daysSinceCreation < 21 ? daysSinceCreation : 20;
```
addison-grant, line 287:
```suggestion
      const numDots = Math.min(daysSinceCreation, 20);
```

Alternative suggestion. A little more concise, and I think as readable.

GrantMatejka:
CR :corn:

erinemay:
QA 🦄 
Looks good. 
 - Tooltip pop-over displays actual day count as opposed to the dot (possibly truncated) count ✅ 
 - All other Tooltip pop-overs continue to behave as expected ✅ 
 - After seeing the different yellow colors over the past commits, this is the best for both distinguishing between the yellow dots and the white background in light mode AND distinguishing between yellow and orange. ✅ 


DEPLOY_BLOCK🌵  on a note/question

I agree that the dot count need to be truncated. Some cards look ridiculous when the dot count = day count.

I see the age_container element extends to the full width of the card, but the dots don't extend to the full width of the card. 
It does not feel intuitive to me that this represents full/overflowing when there is still a lot of blank space after the dots end. 

image

I tried narrowing the screen to see when the age indicator dots would start wrapping and  it looks like other elements will start to wrap before them on most pulls. 

Screen Shot 2021-04-30 at 9 23 58 AM

image

So my question is: Is there a better number to stop the indicator on that will make it the appearance a little nicer? 
If 20 was chosen for reasons other than display and those reasons are more important than a little appearance issue, then I'm good to deploy as is. As is, it serves the purpose for which it was created: To clearly indicate age of a pull. 
CC: @MilesRobinson

GrantMatejka:
CR :corn:

erinemay:
QA 👍 
 - New Tooltip pop-over location looks better (in line with the indicator)
 - The capped-at-30 appearance is a lot more pleasing than 20 (indicator bar of dots appears "full" at max count)
 - When narrowing the Pulldasher page, the display columns re-arrange (from 3 to 2 columns wide) before the dots begin to wrap (at different zoom values)
 - Hovering over the dots triggers the Tooltip pop-over as opposed to hovering over the entire section.

cdcline:
CR 👍

mlahargou:
deploy_block 📱  @sterlinghirsh do you have any recommendations on color changes?

sterlinghirsh:
I liked the idea of using a single emoji like 👶 🧒 🧑 🧓 👴 💀 . Since we're using r/y/g for test-status and blue and purple for CR/QA status, that doesn't give us a ton of options without getting very rainbow. We're not using pink, cyan, or chartreuse anywhere, but I think if we're hard set on the idea of using a line of dots to indicate age, make them gray or some other neutral color. Age of pull shouldn't be visually highlighted over the CI/CR/QA status.

sterlinghirsh:
Proposed mappings:
- 👶 < 1 hours
- 🧒 < 12 hours (i.e. this workday)
- 🧑 < 2 days
- 🧓 < 7 days
- 👴 < 14 days
- 💀 all others

For me, the 🧓 and 👴 look a little different, with the latter being bald, which makes him look older, but I'm not sure if that's true for everyone else's emoji display.

sterlinghirsh:
I like the addition of the pull age indicator, but it would be nice if we don't make the pull-cards any/much taller and don't add too much visual clutter.

MilesRobinson:
I disagree with the use of emojis.  Emojis require a cognitive translation layer. A count is a count is a count.  The current pips being displayed communicates that very simply with minimal real estate impact. If we do not like the colors for conflicting of "state" meanings, then a simple #999999 would be sufficient.  But the very point of it is to have an easy to understand metric that demonstrates the reality on the ground of how long our pulls take.  In looking at the dev example of pulldasher, there is a rewarding feeling (at least for me) from seeing a list of issues in the QA and CR columns with only 1 or 2 pips in them.  And it is an immediate contrast for issues that could use a bit of lovin.  

My issue with the emoji solution is that these automatic visual interpretations get lost when looking at the page as a whole, and remove the ability to scan the data immediately.  

The intended value goals here are:
1) to include pull aging as a consideration in addressing pulls in pull dasher
2) to know the time the pull has been open
3) to be able to quickly and easily assess the age of pulls relative to others
4) to be able to quickly and easily assess the overall responsiveness of our pull management organically in pull dasher rather than requiring some other dashboard or chart (because that requires leaving the flow where the data is most relevant for decision making on what to take up next).

I think that emojis are not distinct enough to distinguish at a glance, nor to do they provide immediate relative context without some mental translation and training to understand and interpret.  And they would be visually hard to group when evaluating the board as a whole.

sterlinghirsh:
Fair points @MilesRobinson . I'm curious to see it with emojis if it's easy but I'm not married to it. I think you're right that in general, icons do have an additional cognitive layer, though I think faces are a potential exception since we have a dedicated part of our brains (fusiform gyrus) for recognizing faces, so we may actually see a benefit from conveying this info by face icons. Of course many things play into that, including the specifics of the emoji set, how big / detailed they are, etc, so maybe it wouldn't actually be better in this case. Still, I think about the game Guess Who, which involves scanning a field of faces and grouping them by characteristics, which comes naturally even for children.

mlahargou:
Are we okay to un_deploy_block this? I personally don't have a problem with the color scheme and it seems as if QA and CR was happy with it.

addison-grant:
I think we can deploy as-is and think about revamping the look and feel some more post-deploy.

mlahargou:
👍 

un_deploy_block 📱

sterlinghirsh:
The colors are too much. Gray would be nice. If you have to use r/y/g please use the same colors we're already using rather than full-bright red etc.

mlahargou:
I can make them the same.

Issue #227, opened 2021-12-10
Add support for GitHub Actions


mlahargou:
TIL that GitHub does not treat Actions on commits as **status**es. Instead, they are treated as **checks**.

We need to add support for **checks** by merging the statuses and checks in [`getCommitStatuses()`](https://github.com/iFixit/pulldasher/blob/52d2ef4d926d56e2f5f28bd42dd3fd5b57bea351/lib/git-manager.js#L388-L393)

I'm thinking we do something like:
```
function getCommitStatuses(repo, ref) {
   debug("Getting commit status for %s", ref);
   let statuses = githubRest.repos.getCombinedStatusForRef(params({ ref }, repo))
      .then(res => res.data.statuses)
      .then(statuses => statuses || [])
   let checks = githubRest.checks.listForRef(params({ ref }, repo))
      ....
   return statuses.concat(checks);
}
```

Then, when we try to create the status objects, we can pull off the relevant information [here](https://github.com/iFixit/pulldasher/blob/12a29dd1f636aec1c80a2ced45c24eee02926d27/lib/git-manager.js#L206-L214):
```
         let statuses = commitStatuses.map(function(commitStatus) {
            return new Status({
               repo:          repo,
               sha:           githubPull.head.sha,
               state:         commitStatus.state || commitStatus.conclusion || commitStatus.status,
               description:   commitStatus.description || commitStatus.conclusion || commitStatus.status,
               target_url:    commitStatus.target_url || commitStatus.html_url, 
               context:       commitStatus.context || commitStatus.name,
            });
         });
```

CC @davidrans @danielbeardsley @BaseInfinity

mlahargou:
For checks:
- `commitStatus.conclusion`:
  - `null` until the check has finished
  - Can be one of: `action_required`, `cancelled`, `failure`, `neutral`, `success`, `skipped`, `stale`, or `timed_out`
- `commitStatus.status`"
  - Never null
  - Can be one of: `queued`, `in_progress`, or `completed`

PR #224, opened 2021-10-29
New Frontend: Add pull card and update tooling

danielbeardsley:
More info in each commit.

* Update tooling
   * New eslint
   * Make eslint aware of typescript
   * New jsx/tsx tranform (no longer need `import React ...` in every file)
* Add functionality
   * Store `Pull` class in state and replace each when updated
   * Add Pull Card (with commit status UI)

New react-based UI as of now:
image

PR #228, opened 2021-12-14
Pulldasher: Add support for GitHub Actions

mlahargou:
This adds support for GitHub Actions. Normal GitHub status checks should remain unchanged.

We turned on the check_run webhook on the ifixit org. We now listen for that event and create 
a dbStatus which is entered into the DB the same way that commit statuses are.

The manual refresh works by pinging the github api for a given sha. We had to concatenate both
the statuses and the checks. We then pull off the relevant information and send that to the
`commit_statuses` table.

CC @danielbeardsley @BaseInfinity 

Closes https://github.com/iFixit/pulldasher/issues/227

### DIFF
--- a/lib/db-manager.js
+++ b/lib/db-manager.js
@@ -137,7 +137,7 @@ var dbManager = module.exports = {
       let repo = body.repository.full_name;
       let sha = body.check_run.head_sha;
       let state = utils.mapCheckToStatus(body.check_run.conclusion || body.check_run.status);
-      let desc = utils.mapCheckToStatus(body.check_run.conclusion || body.check_run.status);
+      let desc = state;
       let url = body.check_run.html_url;
       let context = body.check_run.name;
 
--- a/lib/db-manager.js
+++ b/lib/db-manager.js
@@ -1,5 +1,6 @@
 var _ = require('underscore'),
     debug = require('./debug')('pulldasher:db-manager'),
+    utils = require('./utils'),
     Promise = require('bluebird'),
     db = require('../lib/db'),
     DBIssue = require('../models/db_issue'),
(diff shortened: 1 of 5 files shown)


jarstelfox:
The errors look like a node version issue: syntax used in some dependency is newer than node can handle

mlahargou:
Looks like ci hasn't been running on this repo since June(ish). @danielbeardsley, looks like a [pull](https://github.com/iFixit/pulldasher/pull/224) you merged updated quite a few dependencies but didn't run ci. Any chance that PR could have messed up ci?

lib/db-manager.js
```diff
@@ -132,6 +133,41 @@ var dbManager = module.exports = {
       });
    },
 
+   updateCommitCheck: function(body) {
+      let repo = body.repository.full_name;
+      let sha = body.check_run.head_sha;
+      let state = utils.mapCheckToStatus(body.check_run.conclusion || body.check_run.status);
+      let desc = utils.mapCheckToStatus(body.check_run.conclusion || body.check_run.status);
```
danielbeardsley, line 140: Did you intend these to be identical?  maybe `let desc = state` instead so that's explicit?

mlahargou: Yeah. Check runs don't really have a description like status checks do so I just set the description to state.

I can make that change.

mlahargou:
```suggestion
      let desc = state;
```

lib/git-manager.js
```diff
@@ -204,20 +207,43 @@ module.exports = {
...
+               state:         state,
+               description:   desc,
+               target_url:    url,
+               context:       context,
             });
          });
 
+         let checks = commitChecks.map(function(commitCheck) {
+            let state = utils.mapCheckToStatus(commitCheck.conclusion || commitCheck.status);
+            let desc = utils.mapCheckToStatus(commitCheck.conclusion || commitCheck.status);
```
mlahargou, line 227:
```suggestion
            let desc = state;
```

danielbeardsley:
CR :+1: 
Exciting!

Looks good. Maybe we should turn on github actions for pulldasher instead of using Travis. Looks like we're probably running on an older version of node than our tooling (eslint) supports.

mlahargou:
Wait... we don't even have tests on this repo...

danielbeardsley:
> Wait... I don't even think we have tests on this repo...

True, it's just "Build it" for the moment, ensuring there are no compiler errors.

danielbeardsley (approved):
CR :+1:

danielbeardsley:
CR :+1:

mlahargou:
We are going to be switching to GitHub Actions: https://github.com/iFixit/pulldasher/issues/229

jarstelfox:
CR 🌵

ardelato:
QA 👍 

## Steps Taken:
1. Checked out this branch on my cloned Pulldasher directory on Cominor
2. Ran the following script
```
update-pulldasher --run .
```
3. I then tailed the logs from the docker container by running
```
docker logs -f pulldasher
```
4. I then re-ran the **Psalm** GH Actions check for this pull https://github.com/iFixit/ifixit/pull/41026
Re-run job
5. I then confirmed in the logs that the webhook was received and confirmed the new `check_runs` logic was executed by confirming that the new logs were printed
Check Run hook confirmed
Created GH Actions
GitHub Actions Completed

6. I then confirmed on the replica db that the commit status for the latest commit on that PR (https://github.com/iFixit/ifixit/pull/41026/commits/a26a8dfe5a3d1dd74e43d7daef13f72e40acd96c) was updated accordingly
DB Entry Updated

PR #231, opened 2022-01-11
Dockerfile: commit to node 16 for now

danielbeardsley:
Previously, the build broke cause node v17 had a backward incompatible change that broke some crypto thing in a dependency.

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:latest
+FROM node:16-alpine
 
 RUN mkdir -p /opt/pulldasher
 WORKDIR /opt/pulldasher


danielbeardsley:
QA :+1: 

```
▶  docker build .                                                                                                                                                                                                                             
Sending build context to Docker daemon  17.41MB                                                                        
Step 1/8 : FROM node:16-alpine
...
Successfully built 64364a9c19bd
```

djmetzle (approved):
CR :gear: 

Any concern here over switching from [bullseye to alpine](https://hub.docker.com/_/node)?

djmetzle:
Builds fine now with this patch.
QA :gear:

danielbeardsley:
> Any concern here over switching from [bullseye to alpine](https://hub.docker.com/_/node)?

Good find. I looked all over to try and find what distro `latest` was built on but couldn't.

I have no idea about the diff between the two, but alpine I recognize and bullseye I don't :)

sterlinghirsh:
CR :zap:

PR #241, opened 2022-03-09
Update README.md to test using pulldasher-dev.cominor.com

deltuh-vee:
Just updates the README so that people don't have to try setting up localdev to test Pulldasher pulls.
qa_req 0
cr_req 1


andyg0808:
dev_block :+1: 
This repo is public and open-source; we shouldn't be putting internal information or approaches in the readme here.

deltuh-vee:
>This repo is public and open-source; we shouldn't be putting internal information or approaches in the readme here.

Dang that sucks. I'll close this out then.

PR #242, opened 2022-03-10
commit_statuses: allow null log_url

danielbeardsley:
Some statuses don't have urls, so let's not crash when we get webhooks
about those.

qa_req 0

cr_req 1


andyg0808 (approved):

andyg0808:
CR :goat: through ae18be0

PR #255, opened 2022-03-20
Columns: make Collapsable

### LINKED ISSUE #253: New UI: Collapse columns on click (just like the old UI)
Columns (like CR, QA, Dev Block) are should be collasped when you click the header just like in the old UI.

Bonus points: allow storing and updating state like this via url params so folks can easily make links.

An API like this would be useful.
```javscript
const [collapsed, setCollapsed] = useUrlState('cr', false) // (paramName, defaultValue)
```

danielbeardsley:
Make the individual columns collapsable, and store that state in the url so you can make links to the state and the browser back/forward buttons work.

image

Closes #253

### DIFF
--- a/frontend/src/use-url-state.ts
+++ b/frontend/src/use-url-state.ts
@@ -1,10 +1,11 @@
 import { useState, useCallback, useEffect } from 'react';
 
 type state = string|null;
+type stateSetter = (state: state) => void;
 
 type useUrlStateReturn = [
    string,
-   (state: state) => void,
+   stateSetter,
 ];
 
 /**
--- a/frontend/src/use-url-state.ts
+++ b/frontend/src/use-url-state.ts
@@ -29,11 +29,10 @@ export function useUrlState(paramName: string, paramDefault: string): useUrlStat
 
    // Wrap setState so it pushes history and transforms the url
    const setUrlState = useCallback((newState: state) => {
-      const isDefault = newState === paramDefault;
       // Turn null into the default so state is never null
-      setState(newState === null ? paramDefault : newState);
+      setState(newState || paramDefault);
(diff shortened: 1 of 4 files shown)


frontend/src/use-url-state.ts
```diff
@@ -0,0 +1,85 @@
...
+ * Loads it from a query param on page refresh and updates the url as the state
+ * changes.
+ * paramName is expected not to change over the life of the component.
+ */
+export function useUrlState(paramName: string, paramDefault: string): useUrlStateReturn  {
+   const [state, setState] = useState(paramDefault);
+
+   // Load state from url, only once per component
+   useEffect(() => {
+      watchForPopstate(paramName, setState, paramDefault);
```
andyg0808, line 25: I don't know if it's important, but you might consider making this return a function to remove the event listener, and returning that function from the `useEffect`. It will then be run "before the component leaves the screen" according to the [docs](https://reactjs.org/docs/hooks-reference.html#cleaning-up-an-effect), which allows us to stop getting events that try to set state on a removed React component.

andyg0808 (approved):

andyg0808:
CR :goat: through b92bee9

frontend/src/use-url-state.ts
danielbeardsley: Good point. I didn't realize that happens, but of course it does.

danielbeardsley: Fixed!

jarstelfox:
CR 🌵

hackalot805:
CR 💀

deltuh-vee:
QA 🎬 
Individual columns are collapsible and the state is stored as url parameters.

PR #256, opened 2022-03-21
Cryo Pulls: Hide them and add button to show them

### LINKED ISSUE #252: Cryo Pulls: Hide them in the new UI
### Problem
The old UI hides pulls with the `Cryogenic Storage` label and the new UI doesn't.

### Do This
Let's hide them in the new UI (see `useFilteredPulls`)  and perhaps have a button (or a query param) that shows them.

danielbeardsley:
Add a toggle button in the header to show pulls with the `Cyrogenic Storage` label. Hide them by default.

Also, add a flag for these pulls.

Flag when the label is present:
image

Header button:
image

Closes #252

### DIFF
--- a/frontend/src/navbar.tsx
+++ b/frontend/src/navbar.tsx
@@ -1,6 +1,7 @@
 import { usePulls, useSetFilter } from './pulldasher/pulls-context';
 import { Pull } from './pull';
 import { Button, HStack, Center, Flex, Box, BoxProps, Input } from "@chakra-ui/react";
+import { useEffect, useCallback } from "react";
 import { useBoolUrlState } from "./use-url-state";
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faSnowflake } from '@fortawesome/free-solid-svg-icons'
@@ -19,12 +20,8 @@ export function Navbar(props: BoxProps) {
          return patterns.every((pattern) => pull.title.match(pattern));
       }: null);
    };
-   const toggleCryoFilter = function() {
-      setShowCryo(!showCryo);
-      setPullFilter('cryo', showCryo ? null : (pull: Pull) => {
-         return !pull.getLabel("Cryogenic Storage");
-      });
-   };
+   const toggleShowCryo = useCallback(() => setShowCryo(!showCryo), [showCryo]);
+   useEffect(() => setPullFilter('cryo', showCryo ? null : isPullCryo), [showCryo]);
 
    return (
       <Center py={2} bgColor="var(--header-background)" {...props}>
@@ -36,7 +33,7 @@ export function Navbar(props: BoxProps) {
                   title="Show pulls with label Cryogenic Storage"
                   colorScheme="blue"
                   variant={showCryo ? 'solid' : 'ghost'}
-                  onClick={toggleCryoFilter}>
+                  onClick={toggleShowCryo}>
                   <FontAwesomeIcon icon={faSnowflake}/>
                </Button>
             </HStack>
@@ -47,4 +44,8 @@ export function Navbar(props: BoxProps) {
          </Flex>
       </Center>
    );
+}
+
+function isPullCryo(pull: Pull): boolean {
+   return !pull.getLabel("Cryogenic Storage");
 }
--- a/frontend/src/navbar.tsx
+++ b/frontend/src/navbar.tsx
@@ -1,6 +1,6 @@
 import { usePulls, useSetFilter } from './pulldasher/pulls-context';
 import { Pull } from './pull';
-import { Center, Flex, Box, BoxProps, Input } from "@chakra-ui/react"
+import { HStack, Center, Flex, Box, BoxProps, Input } from "@chakra-ui/react";
 
 export function Navbar(props: BoxProps) {
    const pulls: Pull[] = usePulls();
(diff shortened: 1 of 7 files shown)


ardelato:
@danielbeardsley I see that this pull includes the same commits that are in `Columns` pull https://github.com/iFixit/pulldasher/pull/255, was the intentionally?

danielbeardsley:
> was the intentionally?

It's built on that branch... but I forgot to force-push something. Now it's all good.

frontend/src/navbar.tsx
```diff
@@ -1,27 +1,49 @@
...
-      setPullFilter(patterns.length ? (pull: Pull) => {
+      setPullFilter('search', patterns.length ? (pull: Pull) => {
          return patterns.every((pattern) => pull.title.match(pattern));
       }: null);
    };
+   const toggleCryoFilter = function() {
+      setShowCryo(!showCryo);
+      setPullFilter('cryo', showCryo ? null : (pull: Pull) => {
+         return !pull.getLabel("Cryogenic Storage");
+      });
```
andyg0808, line 26: Would it make sense to pull the `setPullFilter` out of the `toggleCryoFilter`, so we always set the `cryo` filter, but normally set it to `null`? That would eliminate the need for the backwards-ternary here, where we remove the filter if `showCryo` is `true`, and it would make the value of `showCryo` the thing which controls whether or not the filter is present (right now, the filter isn't triggered by the value of `showCryo` but by this toggle function).

frontend/src/pulldasher/filtered-pulls-state.ts
```diff
@@ -2,24 +2,31 @@ import { useState } from 'react';
...
+            delete filters[filterName];
+         }
+         setTimeout(() => setFilter({...filters}), 0);
       }
    ];
 }
+
+function filterPulls(pulls: Pull[], filters: Filters): Pull[] {
+   return Object.values(filters).reduce((pulls, filter) => pulls.filter(filter), pulls)
+}
```
andyg0808, line 32: That's one of the prettiest `reduce`s I've ever seen!

andyg0808 (approved):

andyg0808:
CR :goat: through 93df43d
deploy_block :+1: on comment.

frontend/src/pulldasher/filtered-pulls-state.ts
```diff
@@ -2,24 +2,31 @@ import { useState } from 'react';
...
-      (filter: FilterFunction) => {
-         setTimeout(() => setFilter({filter: filter || defaultFilter}), 0);
+      filterPulls(pulls, filters),
+      (filterName:string, filter:FilterFunction|null) => {
+         if (filter) {
+            filters[filterName] = filter;
+         } else {
+            delete filters[filterName];
+         }
+         setTimeout(() => setFilter({...filters}), 0);
```
jarstelfox, line 25: This seems ... odd, why are we waiting until we are finished to call `setFilter`? It's a code smell to me. I see you are keeping what was working, but it's worth a poke.

I think you might be able to do something like:

`setFilter(() => {...filters})` to defer it with react

frontend/src/pull-card/flags.tsx
```diff
@@ -3,13 +3,14 @@ import { actionMessage } from '../utils';
 import { Link, Box, useStyleConfig } from "@chakra-ui/react"
 import { IconDefinition } from '@fortawesome/fontawesome-svg-core';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import { faWarning, faMinusCircle, faEye, faEyeSlash } from '@fortawesome/free-solid-svg-icons'
+import { faWarning, faMinusCircle, faEye, faEyeSlash, faIcicles } from '@fortawesome/free-solid-svg-icons'
```
jarstelfox, line 6: I think the snowflake might be more intuitive

jarstelfox:
CR 🌵

frontend/src/navbar.tsx
danielbeardsley: Sorry... I really don't understand what you're saying. We can't really pull `setPullFilter()` out without causing an infinite loop (calling a `setState()` on every render). I suppose we could nest it in `useEffect()` with `showCryo` as a dependency... but I'm not sure of the point, that'd .....

Oh wait... I see the backward ternary you're talking about, because there's a bug! Because we do `setPullFilter()` in here, it acts on the now-out-of-date value of `showCrypto`. Fixing.

danielbeardsley: Fixed. Good advice.

frontend/src/pulldasher/filtered-pulls-state.ts
danielbeardsley: Commit c0adbe4858aaf3866c5a5b3d93fdb43d30fb2273 tells the story here. using `setFitler(() => ...)` was my first instinct too, but it still resulted in it happening in the same callstack as the keyboard event and holding up the main UI thread.  But... I'll try it again in case I got it wrong.

danielbeardsley: Nope, `setState(() => something)` is functional, but it *doesn't* defer the code execution. From my reading of the docs, it's mainly used for when your new state is predictably derived from the previous state: `(bool) => !bool` or something; the docs don't mention it running that function deferred.

We could of course have the *caller* do this, but that felt more like a gotcha.

frontend/src/pull-card/flags.tsx
danielbeardsley: Great call! Fixed!

frontend/src/pulldasher/filtered-pulls-state.ts
jarstelfox: Weird something is certainly off setstate is async: https://javascript.plainenglish.io/react-tips-async-and-setstate-cb539ad62135

jarstelfox:
CR 🌵

frontend/src/pulldasher/filtered-pulls-state.ts
danielbeardsley: Here's profiling erasing a search of `z` .. i.e. one backspace keydown:
image

Without the setTimeout, the whole render takes place in the event handler:
image

danielbeardsley: So, no, `setState()` is not async :shrug: ..   maybe if you call it multiple times? ..... no, not then either.

jarstelfox: Huh super weird. thanks for all the time you spent digging.

andyg0808 (approved):

andyg0808:
CR :goat: through 35ba36c

deltuh-vee:
QA 🎬 
* Cryo pulls are hidden by default
* Cryo pulls can be toggled

PR #262, opened 2022-03-23
Pure components: memoize all the things

danielbeardsley:
React components will always re-render (as in re-run your functional components, not re-create the HTML) and they don't compare props to save re-rendering unless you explicitly use React.memo(). Nearly all our components are pure or at least can depend on the pull updates to replace entire `Pull` structures instead of individual fields.

See: https://stackoverflow.com/questions/53074551/when-should-you-not-use-react-memo/63405621#63405621

* Memoize a bunch of components
* Update search to just set a bool instead of excluding them from the array results in just attribute updates instead of tae

QA :+1:   Make sure the search func si


deltuh-vee:
@danielbeardsley Did you mean to stick a QA label on this or were you adding QA instructions?

cdcline:
CR 👍

frontend/src/pull-card/index.tsx
```diff
@@ -38,9 +39,10 @@ const SigsAndFlags = chakra(Flex, {
    }
 });
 
-export function PullCard({pull}: {pull: Pull}) {
+export const PullCard = memo(
+function PullCard({pull, show}: {pull: Pull, show: boolean}) {
    return (
-      <Card position="relative">
+      <Card position="relative" display={show ? "" : "none"}>
```
Joshua-Miles, line 45: Super nit-picky, but I feel like this is usually a little more intuitive than an empty string? Looks like Chakra was typed to support it.

```suggestion
      <Card position="relative" display={show ? undefined : "none"}>
```

frontend/src/pull-card/index.tsx
```diff
@@ -38,9 +39,10 @@ const SigsAndFlags = chakra(Flex, {
    }
 });
 
-export function PullCard({pull}: {pull: Pull}) {
+export const PullCard = memo(
```
Joshua-Miles, line 42: Also super nit-picky, but could we use arrow functions instead of  the following?
```
PullCard = memo( function PullCard(){
```

It feels a little repetitive- unless this is for react dev-tool analysis or something?

frontend/src/pulldasher/filtered-pulls-state.ts
```diff
@@ -27,6 +27,11 @@ export function useFilteredPullsState(pulls: Pull[]): ReturnType {
...
 
-function filterPulls(pulls: Pull[], filters: Filters): Pull[] {
-   return Object.values(filters).reduce((pulls, filter) => pulls.filter(filter), pulls)
+function markFilteredPulls(pulls: Pull[], filters: Filters): Pull[] {
+   const filterArray = Object.values(filters);
+   const combinedFilter = (pull: Pull): boolean => filterArray.every((filter) => filter(pull));
+   return pulls.map((pull) => {
+      pull.show = combinedFilter(pull);
+      return pull;
+   });
```
Joshua-Miles, line 36: Mutating state like this is normally a no-no- 

I made this declarative by creating a new pull object with a different show property here, which eliminates the need to pass show as a separate prop when rendering a `PullCard`. I have all the code on another branch that I can push if you want to see it.

Some quick tests suggested that it takes approximately twice as long (16ms to 30ms a render), so maybe not worth the performance hit to write it that way?

Not sure how strict we want to be about state immutability; I usually prioritize it when I can.

Joshua-Miles (commented):
Very cool! CR 💙 

deploy_block 💛  on some comments that can be taken with a grain of salt.

frontend/src/column.tsx
```diff
@@ -22,7 +22,8 @@ export function Column(props: ColumnProps) {
          </Flex>
          <Box display={open ? 'block' : 'none'}>
             {props.pulls.map((pull) =>
-               <PullCard key={pull.getKey()} pull={pull}/>
+               // show is passed at top-level so PullCard is still memoizable
+               <PullCard key={pull.getKey()} pull={pull} show={pull.show}/>
```
Joshua-Miles, line 26: This is a little annoying- I appreciate the comment!

May be able to get around passing show as a separate prop by avoiding mutable state, but it might not be worth the performance hit: 
https://github.com/iFixit/pulldasher/pull/262#discussion_r834448510

danielbeardsley:
*Tthank you* for reviewing this @Joshua-Miles !

frontend/src/pull-card/index.tsx
danielbeardsley: Thanks! Good idea.

frontend/src/pull-card/index.tsx
danielbeardsley: > unless this is for react dev-tool analysis or something?

Yes, it's a bit unsightly, though it is only one line. Precisely. Part of the eslint-react plugin is making sure every component has a display name. In fact, this is the recommended pattern for memoized components:
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/display-name.md

frontend/src/pulldasher/filtered-pulls-state.ts
danielbeardsley: > Mutating state like this is normally a no-no-

Agreed. The previous pattern was more straight-forward. We just excluded filtered pulls from the arrays passed to this component (see 29e1548c8ac5cbeff4e5669aab0bc6ea2eb47587). The drag with that approach is that React then removes the entire dom structure for each card.

This state mutation is definitely a cheater approach. Hmm...   But you suggestion is a nice in-between, keeping the `pull.show` pattern, but replacing the top-level pull instance so react re-renders these components without an explicit `show` prop. Something equivalent to: `pull = {...pull, show: false}`.

There's a million ways to skin the cat and I can think of a few more, perhaps more elegant, and still performant. I'll experiment.

Thanks for the nudge!

danielbeardsley:
dev_block :+1:  on trying a different approach that doesn't involve mutable state gotchas.

frontend/src/pulldasher/filtered-pulls-state.ts
Joshua-Miles: My pleasure! Probably not worth spending a ton of time on. Might make the code easier to maintain, but it's still pretty readable as is.

If you give me write access to the repo I'll push my branch up so you can see my implementation.

danielbeardsley:
un_dev_block :+1:  

I went with passing along a `Set` in the context and letting the consumers do a `set.has()` check to get the bool over replacing the `Pull` objects. I didn't like having `pull.show` even exist anyway.

Thoughts on the methodology @Joshua-Miles ?

frontend/src/column.tsx
danielbeardsley: I decided on a different strategy; passing around a Set instead of mutating state.

danielbeardsley:
un_deploy_block :+1:  Comments addressed.

danielbeardsley:
QA :+1: 

Parts of the pull-card don't render if the pull doesn't change:
image

Grey means render was skipped.

cdcline:
CR 👍

jarstelfox:
CR 🌵

PR #263, opened 2022-03-27
Refresh button: add to pull-card

### LINKED ISSUE #258: New UI: Add refresh button for each pull
In the old UI there's a button that tells the backend to "refresh" the pull from Github's API.

### Do This
Add a similar refresh button to the new UI

Old UI (refresh in upper right):
image

danielbeardsley:
The old UI had a button to refresh pulls from the API. Let's add the same thing to the new UI. Also, add a brief highlight when pulls update.

Closes #258

### DIFF
--- a/frontend/src/pull-card/refresh.tsx
+++ b/frontend/src/pull-card/refresh.tsx
@@ -1,0 +1,51 @@
+import { Pull } from '../pull';
+import { refreshPull } from '../backend/pull-socket';
+import { Box, chakra } from "@chakra-ui/react"
+import { useState } from "react"
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faRefresh } from '@fortawesome/free-solid-svg-icons'
+
+const RefreshContainer = chakra(Box, {
+   baseStyle: {
+      lineHeight: 0.9,
+      visibility: "hidden",
+      position: "absolute",
+      cursor: 'pointer',
+      bottom: '3px',
+      right: '3px',
+      borderRadius: '4px',
+      borderWidth: '1px',
+      borderColor: 'var(--refresh)',
+      padding: "5px",
+      background: "white",
+      color: "var(--refresh)",
+      "&.refreshing": {
+         visibility: "visible",
+      },
+      "&:hover": {
+         color: "var(--refresh-hover)",
+         borderColor: "var(--refresh-hover)",
+      },
+      "&:active": {
+         color: "white",
+         backgroundColor: "var(--refresh-hover)",
+      },
+   }
+});
+
+export function RefreshButton({pull}: {pull: Pull}) {
+   const [oldReceivedAt, setOldReceivedAt] = useState<Date|null>(new Date());
+   const waitingOnRefreshing = oldReceivedAt == pull.received_at;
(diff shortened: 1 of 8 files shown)


frontend/src/pull-card/refresh.tsx
```diff
@@ -0,0 +1,51 @@
...
+      visibility: "hidden",
+      position: "absolute",
+      cursor: 'pointer',
+      bottom: '3px',
+      right: '3px',
+      borderRadius: '4px',
+      borderWidth: '1px',
+      borderColor: 'var(--refresh)',
+      padding: "5px",
+      background: "white",
```
ardelato, line 20: Since we are merging #264 into this branch, does this mean that this pull is blocked until #264 gets merged since #264 will override some of these settings? 

https://github.com/iFixit/pulldasher/blob/d1adef54ac2d5781e37569e1219f7c14270d2194/frontend/src/pull-card/refresh.tsx#L20

frontend/src/pull-card/refresh.tsx
```diff
@@ -0,0 +1,51 @@
...
+      color: "var(--refresh)",
+      "&.refreshing": {
+         visibility: "visible",
+      },
+      "&:hover": {
+         color: "var(--refresh-hover)",
+         borderColor: "var(--refresh-hover)",
+      },
+      "&:active": {
+         color: "white",
```
ardelato, line 30: When it comes to setting the colors, do we usually want to opt for colors defined in the `*.less` files?

frontend/src/pull-card/refresh.tsx
```diff
@@ -0,0 +1,51 @@
...
+         backgroundColor: "var(--refresh-hover)",
+      },
+   }
+});
+
+export function RefreshButton({pull}: {pull: Pull}) {
+   const [oldReceivedAt, setOldReceivedAt] = useState<Date|null>(new Date());
+   const waitingOnRefreshing = oldReceivedAt == pull.received_at;
+   const onClick = () => {
```
ardelato, line 39: I am not sure about naming this arrow function that same as the event type, what about renaming it to `handleOnClick`

ardelato:
CR 👍 with questions

frontend/src/pull-card/refresh.tsx
danielbeardsley: Other way around... That one should kinda be blocked on this one. Either way, it shouldn't matter that much, either can be merged first, it'll just make this diff larger though.

frontend/src/pull-card/refresh.tsx
danielbeardsley: Thanks fro prompting, I looked into it this. We could make a new `--refresh-active-color: @white` but It's used *just* during the click operation (i.e. sub-1s) and it's gonna be white in both cases, I'd rather not bloat the theme with unimportant distinctions that don't need to change between dark/light.

ardelato:
CR 👍

evannoronha (approved):
CR 🔥

deltuh-vee:
QA 🎬 
* Pull card is highlighted when refreshing.  
* Button to refresh pull cards works.

PR #264, opened 2022-03-28
Darkmode: OMG, lights out

danielbeardsley:
Add Darkmode. Most of the previous theme was re-used, it didn't require that much tweaking.

I decided to use the chakra native color mode and piggy-back off its class name. We've set so many customer colors that the built in dark theme pretty much only affects the Search box border.

image

New button in the header:
image

### DIFF
--- a/frontend/src/navbar.tsx
+++ b/frontend/src/navbar.tsx
@@ -4,11 +4,12 @@ import { Button, HStack, Center, Flex, Box, BoxProps, Input } from "@chakra-ui/r
 import { useEffect, useCallback } from "react";
 import { useBoolUrlState } from "./use-url-state";
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import { faSnowflake } from '@fortawesome/free-solid-svg-icons'
+import { faSnowflake, faMoon } from '@fortawesome/free-solid-svg-icons'
 
 export function Navbar(props: BoxProps) {
    const pulls: Pull[] = usePulls();
    const setPullFilter = useSetFilter();
+   const [darkmode, setDarkmode] = useBoolUrlState('dark', false);
    const [showCryo, setShowCryo] = useBoolUrlState('cryo', false);
    const updateSearchFilter = function(event: React.ChangeEvent<HTMLInputElement>) {
       const patterns = event.target.value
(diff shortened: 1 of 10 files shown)


frontend/src/navbar.tsx
```diff
@@ -9,7 +9,7 @@ import { faSnowflake, faMoon } from '@fortawesome/free-solid-svg-icons'
 export function Navbar(props: BoxProps) {
    const pulls: Pull[] = usePulls();
    const setPullFilter = useSetFilter();
-   const [darkmode, setDarkmode] = useBoolUrlState('dark', false);
+   const {toggleColorMode} = useColorMode();
```
ardelato, line 12: Look at the Chakra docs it seems they are using a state variable or the usage of `useColorModeValue`, but I am assuming since there is no state variable declared here that Chakra keeps track of the toggled color state?

ardelato:
CR 👍

davidrans:
CR :coffee:

frontend/src/navbar.tsx
danielbeardsley: > Add colorModeManager (Optional, for SSR)

> If you don't use SSR or don't care about this, you don't need to pass anything. Chakra will use localStorageManager by default.

Docs say it's optional and the default uses local storage to keep the state. I concur as that's how it seems to work.

davidrans:
CR :coffee:

danielbeardsley:
deploy_block :+1:  on #263 being merge (this is built on top)

deltuh-vee:
dev_block ✊ 
Pulls don’t seem to be populating.
CleanShot 2022-03-28 at 13 58 35

danielbeardsley:
Could spend some time and try to reproduce that problem with me? It works fine here:
image

danielbeardsley:
un_dev_block :+1: 

Thanks @deltuh-vee for finding a bug!!

deltuh-vee:
QA 🎬 
I can switch back and forth between dark mode.

danielbeardsley:
un_deploy_block :+1:  base pull merged.

davidrans:
CR :coffee:

jarstelfox:
CR 🌵

PR #265, opened 2022-03-28
Backend: load all open pulls from the DB on start

danielbeardsley:
Previously, we only loaded pulls that were in our *repo* array from the config.

This meant that after a restart, some pulls may disappear.
We may receive webhooks for repos not in that array,
show the pulls for a while (cause they are in memory), then they
disappear on restart cause they don't get loaded from the DB.

The repo list is now only used on start when we refresh all pulls from
the API (need to know which repos to ask about).

Future Work
====
We may want to support `{orgname}/*` in the repo-list or something.

### DIFF
--- a/app.js
+++ b/app.js
@@ -13,7 +13,6 @@ var config = require('./lib/config-loader'),
     mainController = require('./controllers/main'),
     hooksController = require('./controllers/githubHooks'),
     reqLogger = require('./lib/debug')('pulldasher:server:request'),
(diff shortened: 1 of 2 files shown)


davidrans:
CR :coffee:

sctice-ifixit:
CR :flags:

danielbeardsley:
QA :+1: 

I deployed this on pulldasher-dev and it seems to work fine!

PR #266, opened 2022-03-29
Frontend: Hide Closed Pulls

danielbeardsley:
The backend won't send us closed pulls in the initial list, but we
hadn't been hiding pulls that *become* closed after the page has been
loaded. This meant that a merged pull would jut sit around in the ready
column, even after it's been closed.

The old UI did this, it was just overlooked.

### DIFF
--- a/frontend/dummy-pulls.json
+++ b/frontend/dummy-pulls.json
@@ -7672,8 +7672,8 @@
   {
     "repo": "iFixit/server-templates",
     "number": 2886,
-    "state": "open",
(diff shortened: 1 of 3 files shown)


danielbeardsley:
QA :+1: 

Tested on pulldasher-dev, pulls that close get disappeared. Also tested locally with the closed dummy-pull.

davidrans:
CR :coffee:

jarstelfox:
CR 🌵

PR #277, opened 2022-04-05
Navbar: Add UI for Repo Filter

### LINKED ISSUE #268: Repo Filter: allow filtering pulls by repo
Requirements
=====
* UI component 
   * multi-select list of repos
   * Selecting *none* of the options reset's to selecting all
   * updates immediately
   * populates list from the open pulls
   * Show the pull count next to each repo name
* Repos
   * list is parsed from the url
   * url is updated when changing the UI component
   * allow repos that don't have pulls to remain in the list if they are parsed from the UI

danielbeardsley:
Add a UI to easily show / hide pulls from individual repos.

Meets all the requirements from the issue:

* UI component 
   * multi-select list of repos
   * Selecting *none* of the options reset's to selecting all
   * updates immediately
   * populates list from the open pulls
   * Show the pull count next to each repo name
* Repos
   * list is parsed from the url
   * url is updated when changing the UI component
   * allow repos that don't have pulls to remain in the list if they are parsed from the URL

Closes #268

### DIFF
--- a/frontend/src/repo-menu.tsx
+++ b/frontend/src/repo-menu.tsx
@@ -1,7 +1,7 @@
 import { useAllPulls, useSetFilter } from './pulldasher/pulls-context';
 import { useArrayUrlState } from './use-url-state';
 import { Pull } from './pull';
-import { Button, Menu, MenuButton, MenuList, MenuDivider, MenuOptionGroup, MenuItemOption } from "@chakra-ui/react";
+import { useConst, Button, Menu, MenuButton, MenuList, MenuDivider, MenuOptionGroup, MenuItemOption } from "@chakra-ui/react";
 import { useEffect, useCallback, useMemo } from 'react'
 
 type RepoCounts = Map<string, number>;
@@ -12,9 +12,16 @@ export function RepoMenu() {
    const [selectedRepos, setSelectedRepos] = useArrayUrlState('repo', []);
    // Nothing selected == show everything, otherwise, it'd be empty
    const showAll = selectedRepos.length === 0;
+   // List from url may contain repos we have no pulls for
+   const urlRepos = useConst(() => new Set(selectedRepos));
    const setPullFilter = useSetFilter();
 
-   const allRepos = useMemo(() => Array.from(new Set(pulls.map((pull) => pull.getRepoName()))), [pulls]);
+   // May include repos frmo the url for which there are no pulls
+   const allRepos = useMemo(() => {
+      // All repos of open pulls
+      const pullRepos = new Set(pulls.map((pull) => pull.getRepoName()));
+      return [...new Set([...pullRepos, ...urlRepos])]
+   }, [pulls]);
    const repoToPullCount = useMemo(() => getRepoToPullCount(pulls), [pulls]);
 
    const onSelectedChange = useCallback((newSelectedRepos: string | string[]) => {
--- a/frontend/src/repo-menu.tsx
+++ b/frontend/src/repo-menu.tsx
@@ -14,7 +14,6 @@ export function RepoMenu() {
    const setPullFilter = useSetFilter();
 
    const allRepos = useMemo(() => Array.from(new Set(pulls.map((pull) => pull.getRepoName()))), [pulls]);
-   const selectedReposSet = useMemo(() => new Set(selectedRepos), [selectedRepos]);
    const repoToPullCount = useMemo(() => getRepoToPullCount(pulls), [pulls]);
 
    const onSelectedChange = useCallback((newSelectedRepos: string | string[]) => {
@@ -51,8 +50,7 @@ export function RepoMenu() {
           {allRepos.map((repo) =>
              <MenuItemOption
                 key={repo}
-                value={repo}
-                isChecked={showAll || selectedReposSet.has(repo)}>
+                value={repo}>
                 {repo} ({repoToPullCount.get(repo) || 0})
              </MenuItemOption>
           )}
--- a/frontend/src/repo-menu.tsx
+++ b/frontend/src/repo-menu.tsx
@@ -1,7 +1,10 @@
 import { useAllPulls, useSetFilter } from './pulldasher/pulls-context';
 import { useArrayUrlState } from './use-url-state';
+import { Pull } from './pull';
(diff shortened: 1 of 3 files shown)


danielbeardsley:
image

deltuh-vee:
dev_block ✊ 
UI component
- [x] multi-select list of repos
- [x] Selecting none of the options reset's to selecting all
- [x] updates immediately
- [ ] populates list from the open pulls
CleanShot 2022-04-06 at 09 47 46
What pull is in this Repo?
- [x] Show the pull count next to each repo name

Repos
- [x] list is parsed from the url
- [x] url is updated when changing the UI component
- [x] allow repos that don't have pulls to remain in the list if they are parsed from the UI

danielbeardsley:
For clarity:

Right now, the counts next to the repos include *all open pulls*. If we want that to be different, we'll have to be specific about what the counts should be. Thoughts @deltuh-vee ?

deltuh-vee:
@danielbeardsley I personally don't find the counts next to the repo to be very helpful, so I'm actually fine with them not being shown.

danielbeardsley:
OK. I think I'll un_dev_block :+1:  then cause I find the counts useful and I can't really think of better values than the ones we have.

deltuh-vee:
QA 🎬 
Repo list is populated from open pulls.

frontend/src/use-url-state.ts
```diff
@@ -18,18 +18,15 @@ type HistoryState = Record<string, string>;
...
-         return;
-      }
-      setState(urlState);
-      return unsubscribe;
+      return urlState === null ? paramDefault : urlState;
+   });
+
+   // Watch for url state changes (back button) only once per component.
+   useEffect(() => {
+      return watchForPopstate(paramName, setState, paramDefault);
```
andyg0808, line 29: Pedantic, but this could be `useEffect(() => watchForPopstate(paramName, setState, paramDefault));` now if you like.

frontend/src/repo-menu.tsx
```diff
@@ -0,0 +1,59 @@
...
+      }
+
+      // Update the url
+      setSelectedRepos(newSelectedRepos);
+
+      // Update the pull filter
+      const selectedReposSet = new Set(newSelectedRepos);
+      setPullFilter('repo', selectedReposSet.size === 0 ? null : (pull) =>
+         selectedReposSet.has(pull.getRepoName())
+      );
```
andyg0808, line 39: I'm betting the fact that there are two places with similar state like this is going to bite us eventually. We may want to move toward allowing the pull filters to be based on the URL state or something down the road.

Redux may be worth considering for some of this, because it provides a global location to store state which can be accessed from multiple places.

frontend/src/repo-menu.tsx
```diff
@@ -49,11 +53,18 @@ export function RepoMenu() {
...
      </MenuList>
    </Menu>
    );
 }
+
+function getRepoToPullCount(pulls: Pull[]): RepoCounts {
+   return pulls.reduce((counts, pull) => {
+      counts.set(pull.getRepoName(), (counts.get(pull.getRepoName()) || 0) + 1);
+      return counts;
+   }, new Map() as RepoCounts);
```
andyg0808, line 79: Might be worth doing this as a [`countBy`](https://lodash.com/docs/4.17.15#countBy), since it makes it clear exactly what we're doing:
```
return _.countBy(pulls, pull => pull.getRepoName());
```
(you'll need to switch the return type to `Record<string, number>` or something like that, I think).

frontend/src/repo-menu.tsx
```diff
@@ -0,0 +1,80 @@
...
+   const pulls = useAllPulls();
+   // Default is empty array that implies show all pulls (no filtering)
+   const [selectedRepos, setSelectedRepos] = useArrayUrlState('repo', []);
+   // Nothing selected == show everything, otherwise, it'd be empty
+   const showAll = selectedRepos.length === 0;
+   // List from url may contain repos we have no pulls for
+   const urlRepos = useConst(() => new Set(selectedRepos));
+   const setPullFilter = useSetFilter();
+
+   // May include repos frmo the url for which there are no pulls
```
andyg0808, line 19:
```suggestion
   // May include repos from the url for which there are no pulls
```

frontend/src/repo-menu.tsx
```diff
@@ -0,0 +1,80 @@
...
+   const allRepos = useMemo(() => {
+      // All repos of open pulls
+      const pullRepos = new Set(pulls.map((pull) => pull.getRepoName()));
+      return [...new Set([...pullRepos, ...urlRepos])]
```
andyg0808, line 23: Depending, might also be worth sorting this array:
```suggestion
      return _.sortBy([...new Set([...pullRepos, ...urlRepos])])
```

or (happens to work because [`sort()` returns the array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)):
```suggestion
      return [...new Set([...pullRepos, ...urlRepos])].sort();
```

andyg0808 (approved):

andyg0808:
CR :goat: through b8807e8
deploy_block :+1: on optional comment.

frontend/src/repo-menu.tsx
danielbeardsley: My first reaction was "Yes, url is the state, that's how it was done". So I looked more closely at the code and you are totally right, that's *not* how it's currently done and there's a bug: history navigation is broken!

Fixing!

danielbeardsley: Fixed.

frontend/src/repo-menu.tsx
danielbeardsley: Fixed in other pull.

frontend/src/repo-menu.tsx
danielbeardsley: Good idea, fixed!

frontend/src/use-url-state.ts
danielbeardsley: Fixed!

danielbeardsley:
Great CR @andyg0808 !!

Since the author-filter pull refactors this code significantly, I've addressed your concerns there. See the recent commits at #285 

I think I'm going to close this pull and just let the other one do the master merge.

PR #288, opened 2022-04-08
QA Column: exclude pulls failing CI

### LINKED ISSUE #282: React UI: Exclude failed CI pulls from QA column
In the new React UI, pulls that have failed CI checks are showing up in the QA column. They should be excluded.

danielbeardsley:
Generally, if a pull is failing CI, we'll need to do more work on it
and QA should spend there energy elsewhere.

So, we'll hide pulls that fail CI from the QA Column.

Closes #282

### DIFF
--- a/frontend/src/pulldasher/index.tsx
+++ b/frontend/src/pulldasher/index.tsx
@@ -11,7 +11,7 @@ export const Pulldasher: React.FC = function() {
    const pullsReady = pulls.filter(pull => pull.isReady() && pull.isCiRequired());
    const pullsDevBlocked = pulls.filter(pull => pull.getDevBlock());
    const pullsNeedingCR = pulls.filter(pull => !pull.isCrDone() && !pull.getDevBlock());
-   const pullsNeedingQA = pulls.filter(pull => !pull.isQaDone() && !pull.getDevBlock());
+   const pullsNeedingQA = pulls.filter(pull => !pull.isQaDone() && !pull.getDevBlock() && pull.hasPassedCI());
    const leadersCR = getLeaders(pulls, (pull) => pull.cr_signatures.current);
    const leadersQA = getLeaders(pulls, (pull) => pull.qa_signatures.current);
    return (<>


frontend/src/pulldasher/index.tsx
```diff
@@ -11,7 +11,7 @@ export const Pulldasher: React.FC = function() {
    const pullsReady = pulls.filter(pull => pull.isReady() && pull.isCiRequired());
    const pullsDevBlocked = pulls.filter(pull => pull.getDevBlock());
    const pullsNeedingCR = pulls.filter(pull => !pull.isCrDone() && !pull.getDevBlock());
-   const pullsNeedingQA = pulls.filter(pull => !pull.isQaDone() && !pull.getDevBlock());
+   const pullsNeedingQA = pulls.filter(pull => !pull.isQaDone() && !pull.getDevBlock() && pull.hasPassedCI());
```
ardelato, line 14: I am not sure if I am missing something when looking at how `hasPassedCI` works, but won't it return false if there are no build statuses or CI for the pull?

https://github.com/iFixit/pulldasher/blob/d6fc41cf37a86fd38e69ef8684e10b6dcf18606c/frontend/src/pull.ts#L66-L72

https://github.com/iFixit/pulldasher/blob/d6fc41cf37a86fd38e69ef8684e10b6dcf18606c/frontend/src/pull.ts#L113-L118

danielbeardsley: `[].every(() => false)` is true.  i.e. an empty array will *always* return true for `.every()`

image

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

ardelato: Ooooh sick! Thank you @danielbeardsley

ardelato (approved):
CR 👍

deltuh-vee:
QA 🎬 
Pulls that are failing CI are not shown in the QA column.

andyg0808 (approved):

andyg0808:
CR :goat: through 1913dad

PR #290, opened 2022-04-08
Column Headers: show visible pull count

### LINKED ISSUE #269: Column Header pull counts: fix bug
Right now the column headers show the number of pulls in that column as if there were no filters (Cryo, search, repo, ...).

Do This
=======
Make the count match the number of pulls actually shown.

danielbeardsley:
Instead of showing the total pull count (including filtered pulls),
show just the count of pulls the user sees in the columns.

Closes #269

### DIFF
--- a/frontend/src/column.tsx
+++ b/frontend/src/column.tsx
@@ -20,7 +20,7 @@ export function Column(props: ColumnProps) {
          <Flex className="column_header" onClick={() => setOpen(!open)}>
             <Box p={3} pl={4}>{props.title}</Box>
             <Spacer/>
-            <Box className="pull_count" p={3}>{props.pulls.length}</Box>
+            <Box className="pull_count" p={3}>{countPulls(props.pulls, pullsToShow)}</Box>
          </Flex>
          <Box display={open ? 'block' : 'none'}>
             {props.pulls.map((pull) =>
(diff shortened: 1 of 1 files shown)


deltuh-vee:
QA 🎬 
Column counts only count pulls that are shown.

frontend/src/column.tsx
```diff
@@ -20,7 +20,7 @@ export function Column(props: ColumnProps) {
          <Flex className="column_header" onClick={() => setOpen(!open)}>
             <Box p={3} pl={4}>{props.title}</Box>
             <Spacer/>
-            <Box className="pull_count" p={3}>{props.pulls.length}</Box>
+            <Box className="pull_count" p={3}>{countPulls(props.pulls, pullsToShow)}</Box>
```
andyg0808, line 23: It might make sense to move the `pullsToShow` check up into a
```js
const pulls = props.pulls.filter(p => pullsToShow.has(p));
```
above, and then drop the `show` prop below and use
```suggestion
            <Box className="pull_count" p={3}>{pulls.length}</Box>
```
here.

andyg0808 (approved):

andyg0808:
CR :goat: through 2d2d008
deploy_block :+1: on optional suggestion.

frontend/src/column.tsx
danielbeardsley: > and then drop the show prop below and use

The `show` prop is pretty critical to performance: 29e1548c8ac5cbeff4e5669aab0bc6ea2eb47587

danielbeardsley:
un_deploy_block :+1: Natural suggestion, but it's done this way with intent.

ardelato (approved):
CR 👍

jarstelfox:
CR 🌵

PR #300, opened 2022-04-27
CR Leader List: have it show unique CRs in last 14 days

danielbeardsley:
The Leader list was not that helpful. It was showing the total
CR signature count across all pulls in pulldasher's memory (not the DB).
This wasn't that useful of a metric because that was an abritrary number
of pulls and 7 CRs on one pull isn't the same as 1 CR on 7 pulls.

Now:

* Only include open pulls and those merged less than 14 days ago.
* Only include signatures made in the last 14 days
* Count only the number of pulls each user has CRed (not the number of
  signatures)
* Avatars show username on hover

Future work:

* Change pulldasher to load even closed pulls from the DB so
  these numbers stay the same on restart.
   * Right now it only loads open pulls on restart
* Maybe: drop list and only show how you compare to mean or median of the
  group.
  
This is closer to your spec @davidrans

### DIFF
--- a/frontend/src/leader-list.tsx
+++ b/frontend/src/leader-list.tsx
@@ -55,7 +55,7 @@ function colorForIndex(index: number): string {
 export function getLeaders(pulls: Pull[], extractSigsFromPull: (pull: Pull) => Signature[]) {
    const recent = toDateString(new Date(Date.now() - 86400 * 1000 * DAYS_TO_EXAMINE));
    const recentPulls = pulls.filter((pull) => {
-      return pull.merged_at ? pull.merged_at > recent : true;
+      return pull.closed_at ? pull.closed_at > recent : true;
    });
    const sigCounts = new Map<number, SigCount>();
    recentPulls.forEach((pull) => {
--- a/frontend/src/leader-list.tsx
+++ b/frontend/src/leader-list.tsx
@@ -22,6 +22,7 @@ export function LeaderList({leaders, title}: {leaders: SigCount[], title: string
          <chakra.span mb={1} flexShrink={0}>{title}:</chakra.span>
          {leaders.map(({user, count}, i) =>
             <HStack
+               title={user.login}
                overflow="hidden"
                borderRadius="4px"
                fontWeight="bold"
--- a/frontend/src/leader-list.tsx
+++ b/frontend/src/leader-list.tsx
@@ -1,4 +1,5 @@
 import { Pull } from './pull';
+import { toDateString } from './utils';
 import { Signature, SignatureUser } from './types';
 import { Flex, HStack, Box, chakra } from "@chakra-ui/react"
 
@@ -6,6 +7,8 @@ interface SigCount {
    user: SignatureUser;
    count: number;
 }
+
+const DAYS_TO_EXAMINE = 14;
 
 export function LeaderList({leaders, title}: {leaders: SigCount[], title: string}) {
    return (
@@ -49,16 +52,27 @@ function colorForIndex(index: number): string {
 }
 
 export function getLeaders(pulls: Pull[], extractSigsFromPull: (pull: Pull) => Signature[]) {
-   const signatures: Signature[] = pulls.flatMap(extractSigsFromPull);
-   const users = signatures.map((signature) => signature.data.user);
-   const groupedByUsers = users
-      .reduce((grouped: Map<number, SigCount>, user: SignatureUser) => {
-         const group = grouped.get(user.id) || {user, count:0};
-         group.count++;
-         grouped.set(user.id, group)
-         return grouped;
-      }, new Map())
-      .values();
-   return Array.from(groupedByUsers)
+   const recent = toDateString(new Date(Date.now() - 86400 * 1000 * DAYS_TO_EXAMINE));
+   const recentPulls = pulls.filter((pull) => {
+      return pull.merged_at ? pull.merged_at > recent : true;
+   });
+   const sigCounts = new Map<number, SigCount>();
+   recentPulls.forEach((pull) => {
+      const sigs: Signature[] = extractSigsFromPull(pull);
+      const recentSigs = sigs.filter((sig) => sig.data.created_at > recent);
+      // Count one sig per user from this pull
+      const seenUsers = new Set<number>();
+      recentSigs.forEach((sig) => {
+         const user = sig.data.user;
+         if (seenUsers.has(user.id)) {
+            return;
+         }
+         seenUsers.add(user.id);
+         const sigCount = sigCounts.get(user.id) || {user, count:0};
+         sigCount.count++;
+         sigCounts.set(user.id, sigCount)
+      });
+   });
+   return Array.from(sigCounts.values())
       .sort((a: SigCount, b: SigCount) => b.count - a.count);
 }
--- a/frontend/src/types.ts
+++ b/frontend/src/types.ts
@@ -48,7 +48,7 @@ export interface Signature {
       number: number;
       user: SignatureUser;
       type: SignatureType;
-      created_at: DateString | null;
+      created_at: DateString;
       active: number;
       comment_id: number;
       source_type: CommentSource;
--- a/frontend/src/utils.ts
+++ b/frontend/src/utils.ts
@@ -13,3 +13,7 @@ function formatDate(date: DateString) {
    });
 }
 
+export function toDateString(date: Date): DateString {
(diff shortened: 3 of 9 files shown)


HCutrone (approved):
CR :cow2:

bhuminson (approved):
CR :dagger:

jordycosta:
**Current:**
current

**Dev:**
dev

The number of inactive CR's are now being included, is this intended?

jordycosta:
(in progress) 🟢 

Avatars show username on hover ☑️ 
Count # of unique CR stamps ☑️ 
Limit signatures to 14 days ☑️ 

Count for only open pulls & pulls merged in the past 14 days*
_I can say for sure that open pulls are being counted, but it doesn't appear that merged pull signatures are

---

**No longer shown in CR column:**
date

label

frontend/src/leader-list.tsx
```diff
@@ -49,16 +53,27 @@ function colorForIndex(index: number): string {
...
-         const group = grouped.get(user.id) || {user, count:0};
-         group.count++;
-         grouped.set(user.id, group)
-         return grouped;
-      }, new Map())
-      .values();
-   return Array.from(groupedByUsers)
+   const recent = toDateString(new Date(Date.now() - 86400 * 1000 * DAYS_TO_EXAMINE));
+   const recentPulls = pulls.filter((pull) => {
+      return pull.merged_at ? pull.merged_at > recent : true;
```
evannoronha, line 58: Minor point, but I'm reading this as: Look at all PRs that were merged in the past 14 days, or that do not have a merged_at value. I think this means that the leaderboard will also take into account CR stamps on open PRs that have stuck around longer than 14 days.

evannoronha (approved):
CR 🔥

frontend/src/leader-list.tsx
danielbeardsley: > . I think this means that the leaderboard will also take into account CR stamps on open PRs that have stuck around longer than 14 days.

You are correct. That was intentional. We mostly want "all recent CRs" regardless of when the pull was opened / merged. I just decided to trim the old merged pulls for performance.

danielbeardsley:
So, I went for it. This now keeps recently merged pulls in memory, sends them to the client, and culls too-old ones every hour.

Looking good!
image

frontend/src/leader-list.tsx
```diff
@@ -49,16 +52,27 @@ function colorForIndex(index: number): string {
...
-   const groupedByUsers = users
-      .reduce((grouped: Map<number, SigCount>, user: SignatureUser) => {
-         const group = grouped.get(user.id) || {user, count:0};
-         group.count++;
-         grouped.set(user.id, group)
-         return grouped;
-      }, new Map())
-      .values();
-   return Array.from(groupedByUsers)
+   const recent = toDateString(new Date(Date.now() - 86400 * 1000 * DAYS_TO_EXAMINE));
```
ardelato, line 56: I'm confused by the different date formats we are using. In our DB we store dates in seconds, but for Pulldasher we are using milliseconds?

lib/pull-manager.js
```diff
@@ -77,3 +78,13 @@ pullQueue.on('pullsChanged', function(pulls) {
...
    });
 });
+
+// Cull old closed pulls from memory every once in a while
+setTimeout(() => {
+   const oldestAllowed = pullManager.getOldestAllowedPullTimestamp() * 1000;
+   pulls = pulls.filter((pull) =>
+      pull.isOpen() ||
+      !pull.data.closed_at ||
+      pull.data.closed_at.getTime() > oldestAllowed
```
ardelato, line 88: I'm confused to how `closed_at` has the function `getTime()` since it looks like `pull.data` is of type `PullData` which I then see has `DateString` as the type for `closed_at`?

closed_at

lib/pull-manager.js
```diff
@@ -8,6 +8,12 @@ var sockets = [];
 var pulls = [];
 
 var pullManager = module.exports = {
+   getOldestAllowedPullTimestamp: function() {
```
ardelato, line 11: Can we rename this function to `getOldestAllowedPullTimestampInSeconds` ? I see we are using it to represent Unix timestamps but we are working with both `seconds` and `milliseconds` in other parts of the code which was making it hard to decipher which format we are trying to use.

frontend/src/utils.ts
```diff
@@ -13,3 +13,7 @@ function formatDate(date: DateString) {
    });
 }
 
+export function toDateString(date: Date): DateString {
```
ardelato, line 16: We may want to move over some of the util functions from `pulldasher/lib/utils.js` as there seems to be more functions that deal with dates in there as well https://github.com/iFixit/pulldasher/blob/923c28d9a8a1c13d3d5e8abca03d2058fd96b30d/lib/utils.js#L34-L36

lib/pull-manager.js
danielbeardsley: Sorry for the confusion. All code outside of `frontend/` is running on the backend. Thus it has different models / types / no typescript. This is a `models/pulls.js` object and thus `pull.data.closed_at` is a `DateTime`.

lib/pull-manager.js
danielbeardsley: I understand the confusion. Maybe better to have this return a `Date` object and the caller can do what it needs.

danielbeardsley: Done!

frontend/src/utils.ts
danielbeardsley: Yes, there are more date functions we should copy them as needed. At the moment, the backend doesn't support typescript, so we can't import them directly.

danielbeardsley:
Thanks @ardelato for looking at this and pointing out an improvement!

k0rvusk0r4x:
- Count with avatars appears correctly.
- Name shows on hover.
- Shows count of unique CR's.
- Counts stamps of open and up to last fourteen days.

QA 😺

frontend/src/leader-list.tsx
```diff
@@ -49,16 +53,27 @@ function colorForIndex(index: number): string {
...
+         if (seenUsers.has(user.id)) {
+            return;
+         }
+         seenUsers.add(user.id);
+         const sigCount = sigCounts.get(user.id) || {user, count:0};
+         sigCount.count++;
+         sigCounts.set(user.id, sigCount)
+      });
+   });
+   return Array.from(sigCounts.values())
```
andyg0808, line 77: An alternative. Not sure if it's easier to read or not. It's shorter and uses no mutation.
Note that `countBy` and `uniq` are Lodash functions.

```suggestion
   // Generate map from userid to signatures (valid & one-per-PR)
   const sigGroups: Record<number, SignatureUser> = groupBy(recentPulls.flatMap((pull) => {
      const sigs: Signature[] = extractSigsFromPull(pull);
      const recentSigs = sigs.filter((sig) => sig.data.created_at > recent);
      // Only include each user once per PR
      return uniq(recentSigs.map(sig => sig.data.user));
   }), user => user.id);

   return Object.values(sigGroups).map((arr): SigCount => ({
      user: arr[0],
      count: arr.length,
   }));
```

andyg0808 (approved):

andyg0808:
CR :goat: through 3464c78

Cool!

ardelato (approved):
CR 👍

frontend/src/leader-list.tsx
danielbeardsley: Unfortunately it seems there's a bug. Each `sig.data.user`  is `{id: number, login: string}` thus using `uniq` on an array of objects won't do anything (I think) cause they are all unique objects.

Changing that `uniq` line would probably work, but I find that harder to follow. Maybe a few more variable assignment steps would help. But to my eye it's easier to validate the correctness of the original code.

andyg0808: Oh, rats. I'd originally written it with `sig.data.user.id`, which of course is much better behaved. I'm fine with the way you've got it, since you like that better.

frontend/src/types.ts
```diff
@@ -48,7 +48,7 @@ export interface Signature {
       number: number;
       user: SignatureUser;
       type: SignatureType;
-      created_at: DateString | null;
```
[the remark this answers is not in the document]
danielbeardsley: Correct, changing this doesn't change reality. But in this case, this is behind reality:
3bc78a18fcbbf51113b846fad42d376b2c746fb6

PR #311, opened 2022-05-04
Search: use native browser search input type

danielbeardsley:
The browser `<input type="search">` has the clear button already, so we can simplify the code.

We can also extract it to a separate component so the navbar component is easier to read.


andyg0808 (approved):

andyg0808:
CR :goat: through 48f787e

PR #315, opened 2022-05-11
Connection state: Expose with icon

danielbeardsley:
Adds an indicator that shows the current connection state of socket.io.

Connected:
image

Reconnecting:
image

PR #319, opened 2022-05-17
Frontend dev: stop socket io connection

danielbeardsley:
We didn't use to try connecting to the socket when only running the frontend in dev mode, but then we added the connection state display #315 and accidentally started attempting a connection.

Here, we fake the connection status when in dev mode. Previously, when running `frontend:start` we were getting one 404 on the client every 10s or so, now we don't.


Michelle5102 (approved):
CR :octocat:

Michelle5102 (approved):
CR :octocat:

danielbeardsley:
QA :+1: 

Before:
image


After:
image

PR #322, opened 2022-05-18
Commit status: tweak bubble sizes

### LINKED ISSUE #321: CI Statuses: Scale bubble size a little better
We're scaling the size of the bubble based on the number of statuses in that state, but the scaling is a bit off, cause 1 shouldn't be the same size as 2 or for that matter 13:

image

danielbeardsley:
The CI bubbles now scale with the number of status in that state. Previously, the relative scaling was unpredictable.

Closes #321 

**Before:**
image

**After:**
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,5 @@
 node_modules
 pulldasher.pid
 dist
+db_data/
 /.env.db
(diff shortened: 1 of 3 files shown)


danielbeardsley:
QA :+1:

davidrans (approved):
CR :coffee:

Issue #239, opened 2022-02-28
Don't invalidate CR or QA stamps for clean master merges


andyg0808:
Provide a configuration point which enables clean master merges to be pushed to a branch without invalidating the existing CR or QA stamps.

This is helpful when working with a repository with the "Require branches to be up to date before merging" setting enabled:
![image](https://user-images.githubusercontent.com/1283490/156054047-2e36ca6d-ce93-4cb8-8c21-e83d52f2bd24.png)

PR #324, opened 2022-05-28
Do not invalidate signatures on clean merge

josmfred:
This was some functionality that we needed for our instance of pulldasher, so I added a quick solution.

One peculiarity is, since selecting the current changes during a conflict is an empty change, some conflicts resolutions are considered clean.

Thought I would open a pull here in case it could be useful for solving https://github.com/iFixit/pulldasher/issues/239 on your end.

<details>
<summary> With `ignoreCleanMerge` set to `true` </summary>

true
</details>

<details>
<summary> With `ignoreCleanMerge` set to `false` </summary>

false
</details>

Ref: https://github.com/iFixit/pulldasher/issues/239

### DIFF
--- a/lib/git-manager.js
+++ b/lib/git-manager.js
@@ -483,5 +483,5 @@ function filterOutPulls(issues) {
  * Extract the commit date from commit object
 */
 function getCommitDate(commit) {
-   return commit.commit.committer.date;
+   return new Date(commit.commit.committer.date);
 }
--- a/lib/git-manager.js
+++ b/lib/git-manager.js
@@ -172,11 +172,11 @@ module.exports = {
          }, commentSignatures);
 
          // Signoffs from before the most recent commit are no longer active.
-         var latestCommitDate = new Date(latestCommit.commit.committer.date);
+         var latestDate = new Date(latestCommit ? latestCommit.commit.committer.date : githubPull.created_at);
          signatures.forEach(function(signature) {
             if ((signature.data.type === 'CR' ||
              signature.data.type === 'QA') &&
-             new Date(signature.data.created_at) < latestCommitDate) {
+             new Date(signature.data.created_at) < latestDate) {
                signature.data.active = false;
             }
          });
--- a/lib/git-manager.js
+++ b/lib/git-manager.js
@@ -128,7 +128,7 @@ module.exports = {
 
       var reviewComments = getPullReviewComments(repo, githubPull.number);
       var comments = getIssueComments(repo, githubPull.number);
-      var headCommit = getCommit(repo, githubPull.head.sha);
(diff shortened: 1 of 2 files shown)


lib/git-manager.js
```diff
@@ -411,6 +411,48 @@ function getCommit(repo, sha) {
...
+      .then(commits => {
+         return Promise.all(commits.map(commit => {
+            if (shouldFetchFiles(commit)) {
+               return getCommit(repo, commit.sha);
+            } else {
+               return Promise.resolve(commit);
+            }
+         }))
+      })
+      .then(expandedCommits => {
```
danielbeardsley, line 445: At this point, we have the all the commits for the whole pull, But the order seems unspecified. The API doesn't say anything about the order of commits, but I don't think there's anything we can intrinsically order them by. It seems like the rest of the code expects them to be in oldest-first order.

danielbeardsley, line 445: At this point, we have the all the commits for the whole pull, But the order seems unspecified. The API doesn't say anything about the order of commits, but I don't think there's anything we can intrinsically order them by. It seems like the rest of the code expects them to be in oldest-first order.

danielbeardsley:
This looks good! So, CR :+1:  on this part, but it's only a partial solution. This code change covers us when *refreshing* the pull (I think). When we get notified via a webhook that a new commit is pushed, we immediately invalidate the signatures:

https://github.com/iFixit/pulldasher/blob/5f07ecbc5fc9ca3bd85e15f977836d1bb1870b04/controllers/githubHooks.js#L65-L69

We could solve this systematically (harder) by not storing `active` and instead storing the HEAD commit at the time of signature creation (or something). But then that's harder to re-create after the fact.

danielbeardsley:
This looks good! So, CR :+1:  on this part, but it's only a partial solution. This code change covers us when *refreshing* the pull (I think). When we get notified via a webhook that a new commit is pushed, we immediately invalidate the signatures:

https://github.com/iFixit/pulldasher/blob/5f07ecbc5fc9ca3bd85e15f977836d1bb1870b04/controllers/githubHooks.js#L65-L69

We could solve this systematically (harder) by not storing `active` and instead storing the HEAD commit at the time of signature creation (or something). But then that's harder to re-create after the fact.

lib/git-manager.js
josmfred: Duplicate of https://github.com/iFixit/pulldasher/pull/324#discussion_r885005330

josmfred: I believe I read on a post somewhere that they were chronologically ordered by commit date by default, but it definitely does not guarantee it in the api documentation. I will add a change to explicitly check this.

[Here](https://github.blog/changelog/2020-07-14-pull-request-commits-now-ordered-chronologically/) is a blog post about it, but we shouldn't rely on that behavior.

Issue #325, opened 2022-06-10
Pulldasher: Support mobile


jarstelfox:
![62E2C93C-0AE7-4CE6-B2FC-15A22D9BF7D7](https://user-images.githubusercontent.com/857362/172978810-17c033fb-dfc8-4d5e-a976-db1f957a1111.png)

😅

andyg0808:
![image](https://user-images.githubusercontent.com/1283490/190237300-21948142-8b6c-4aa6-86b7-34e1e2e056c9.png)

PR #326, opened 2022-06-20
Mobile support: MVP

danielbeardsley:
Add just a dash of mobile support:

* Set a min-width on the columns so the format goes to 2-wide, then 1-wide as it gets smaller
* Hide some header widgets as browser size gets smaller so it's not too crowded

This gets us most of the way there without too much effort.

Connect #325


andyg0808 (approved):

andyg0808:
CR :goat: through 2fab5ca

rickisXP:
QA :+1: 
- Pull cards on mobile viewports look fine and are not overcrowded :heavy_check_mark: 
- Functionality remains consistent with desktop version :heavy_check_mark:

PR #327, opened 2022-07-11
Add notifications for pulls which are ready for actions

andyg0808:
This notifies the current user when there are pulls for them to re-review or merge.


frontend/src/notifications.tsx
```diff
@@ -0,0 +1,50 @@
...
+) {
+  const [seen, setSeen] = React.useState<KeyType[]>([]);
+  const unseen = xs.filter((x) => !seen.includes(key(x)));
+  const filtered = unseen.filter(filter);
+  React.useEffect(() => {
+    if (filtered.length > 0) {
+      const msg = message(filtered);
+      if (Notification.permission === "granted") {
+        new Notification(msg);
+        setSeen((seen) => [...seen, ...filtered.map(key)]);
```
danielbeardsley, line 26: This means it will only ever notify about a given pull once per page load, even if that pull goes back and forth between needing review and being reviewed. Am I reading this correctly? If so, is that what we want?

danielbeardsley:
This is really cool!

frontend/src/notifications.tsx
andyg0808: Oh, good point. I don't think we want that; I've got an idea.

andyg0808:
Sorry about the force-push; it only changed the second-to-last commit.

frontend/src/notifications.tsx
andyg0808: Okay, fixed in [2f6557c](https://github.com/iFixit/pulldasher/pull/327/commits/2f6557c1b7fa8469195220cf072502391594c232) and [5cfee99](https://github.com/iFixit/pulldasher/pull/327/commits/5cfee998ee7071ebd9e58bfad986469a4892cd06). Now we always update `seen` to be the list of PRs we _could_ have notified on last time it rendered. That means that if a PR becomes ineligible to be notified about it'll get dropped from the list, which means it'll get notified for again after it becomes eligible again.

frontend/src/pulldasher/notifications.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { getUser } from "../page-context";
+import { Pull } from "../pull";
+
+export function usePullNotification(pulls: Pull[], action: string) {
+  useNotification<Pull>(pulls, {
+    key: (pull: Pull) => pull.number,
+    filter: (pull: Pull) => pull.isMine() || pull.hasOutdatedSig(getUser()),
+    message(pulls) {
+      const titles = pulls.map((p) => p.title).join(", ");
+      console.log(action, titles);
```
jarstelfox, line 11:
```suggestion
```

frontend/src/pulldasher/notifications.tsx
```diff
@@ -0,0 +1,18 @@
...
+      if (pulls.length === 1) {
+        return `Pull ready for ${action}: ` + titles;
+      }
+      return `Pulls ready for ${action}: ${titles}`;
```
jarstelfox, line 15:
```suggestion
      const pull = pulls.length === 1 ? 'Pull' : 'Pulls'
      return `${pull} ready for ${action}: ${titles}`;
```

frontend/src/pulldasher/notifications.tsx
jarstelfox, line 11:
```suggestion
      const titles = pulls.map((p) => p.title).join(", ");
      new Audio('https://audio-previews.elements.envatousercontent.com/files/105394911/preview.mp3?response-content-disposition=attachment%3B+filename%3D%22QKTA234-pop.mp3%22').play()
```

https://mixkit.co/free-sound-effects/pop/

jarstelfox (approved):
CR 🌵

jarstelfox (approved):
CR 🌵

andyg0808:
QA :+1: 
Added a bunch of code to allow testing of pull movement; discovered some unintuitive behavior which is now fixed.

frontend/src/pulldasher/notifications.tsx
andyg0808: Dropped in 9456617376afc5226d50d370c176e5c80b0c2528

andyg0808:
Bell from https://commons.wikimedia.org/wiki/File:Bicycle-bell-1.wav (public domain).

andyg0808:
QA :+1: 
Bell works, alerts pop roughly as expected.

frontend/src/pulldasher/notifications.tsx
andyg0808: [ad9d045](https://github.com/iFixit/pulldasher/pull/327/commits/ad9d045242d11c28493d84f37832eb2aea74dfe7)

jarstelfox (approved):
CR 🌵

frontend/src/notifications.tsx
```diff
@@ -0,0 +1,56 @@
...
+  key: (x: T) => KeyType;
+};
+
+export function useNotification<T>(
+  xs: T[],
+  { filter, message, key }: NotificationOptions<T>
+) {
+  const [seen, setSeen] = React.useState<KeyType[]>([]);
+  const filtered = xs.filter(filter);
+  const unseen = filtered.filter((x) => !seen.includes(key(x)));
```
danielbeardsley, line 21: This seems N^2... which is probably fine because needing-QA and needing-CR are rarely bigger than 30 (30^2 == 900). Still a Set might be a better choice.

frontend/src/notifications.tsx
```diff
@@ -0,0 +1,56 @@
...
+  React.useEffect(() => {
+    if (unseen.length > 0) {
+      const msg = message(unseen);
+      if (Notification.permission === "granted") {
+        new Notification(msg);
+        new Audio("/sounds/bell.mp3").play();
```
danielbeardsley, line 27: Woah, notifications don't have their own system-configured sounds?

andyg0808: I don't really know. I didn't notice a sound when it popped up, though.

frontend/src/notifications.tsx
andyg0808: :+1: 
If it becomes a problem, we can deal with it. That analysis is probably still over-pessimistic, because `seen` is `filtered` from the last run, so we're looking at `(filtered.length)^2`. And `filtered` is "pulls I should be notified about," so it should generally be much less than the maximum needing-CR or needing-QA number (like, if I've done ten CRs that are still on Pulldasher, then `filtered` should be ten items long). And 100 sounds totally reasonable to me.

PR #332, opened 2022-08-23
GitHub Actions: Mark skipped jobs as success

mlahargou:
We sometimes skip jobs and want them to not be shown as errors in pulldasher. Let's map `skipped` -> `success` so they don't get stuck in ci_blocked.

Ref: https://ifixit.slack.com/archives/C01DBJK0H7U/p1661213572318169

Closes: https://github.com/iFixit/ifixit/issues/44378

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -66,6 +66,7 @@ module.exports = {
          case 'queued':
             return 'pending';
          case 'success':
+         case 'skipped':
             return 'success';
          case 'failure':
             return 'failure';


ardelato:
QA 🖲️ 

<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td>Before 3@2x
	<td> After 3@2x
</table>

## Steps:

1. Checked out this branch on a cloned `pulldasher` on Cominor
2. Ran `update-pulldasher --run ./`
3. Confirmed for `dependabot` pulls, `diff-cover` and `send-finished` jobs that were skipped were no longer saying they were of status `error` but of `success`

PR #334, opened 2022-09-01
Toggle view blocked pull requests

### LINKED ISSUE #333: Toggle view button for externally blocked PRs
**Do this:**

Create a button with the same behavior as the Cryogenic Storage toggle for pull requests labelled as externally blocked

https://ifixit.slack.com/archives/C90126AN4/p1661978297154469

jordycosta:
### Overview:

firefox_o80hIMmI1g

### CR/QA Notes:

> CR: Review to https://github.com/iFixit/pulldasher/pull/334/commits/5c04bb9da71fbfa6362f70d933506222f98b144a

> QA: https://pulldasher-dev.cominor.com/
>
>I updated Pulldasher-Dev, so you should just be able to click the link above.
>If the changes aren't present, here is a [reference for testing](https://overlord.ifixit.com/Guide/How+to+test+Pulldasher+with+Pulldasher-dev/1124)

cr_req 2
Closes #333

### DIFF
--- a/frontend/src/navbar.tsx
+++ b/frontend/src/navbar.tsx
@@ -18,7 +18,7 @@ import { NotificationRequest } from "./notifications"
 import { useConnectionState, ConnectionState } from "./backend/socket";
 import { useHotkeys } from 'react-hotkeys-hook';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import { faSnowflake, faMoon, faWifi, faCircleNotch, faXmark, faCircleExclamation, faEye, faEyeSlash } from '@fortawesome/free-solid-svg-icons'
+import { faSnowflake, faMoon, faWifi, faCircleNotch, faXmark, faCircleExclamation, faEyeSlash } from '@fortawesome/free-solid-svg-icons'
 
 // Default width of the left and right sections of the nav bar
 const sideWidth = "220px";
@@ -63,9 +63,9 @@ export function Navbar(props: BoxProps) {
                   size="sm"
                   title="Toggle display for externally blocked pull requests"
                   colorScheme="blue"
-                  variant='ghost'
+                  variant={showExtBlocked ? 'ghost' : 'solid'}
                   onClick={toggleShowExtBlocked}>
-                  <FontAwesomeIcon icon={showExtBlocked ? faEyeSlash : faEye}/>
+                  <FontAwesomeIcon icon={faEyeSlash}/>
                </Button>
                <Button
                   display={hideBelowMedium}
--- a/frontend/src/navbar.tsx
+++ b/frontend/src/navbar.tsx
@@ -29,11 +29,15 @@ export function Navbar(props: BoxProps) {
    const setPullFilter = useSetFilter();
    const {toggleColorMode} = useColorMode();
    const [showCryo, setShowCryo] = useBoolUrlState('cryo', false);
+   const [showBlocked, setShowBlocked] = useBoolUrlState('external_block', false);
    const hideBelowMedium = ['none', 'none', 'block'];
    const hideBelowLarge = ['none', 'none', 'none', 'block'];
 
    const toggleShowCryo = useCallback(() => setShowCryo(!showCryo), [showCryo]);
    useEffect(() => setPullFilter('cryo', showCryo ? null : isPullCryo), [showCryo]);
+
+   const toggleShowBlocked = useCallback(() => setShowBlocked(!showBlocked), [showBlocked]);
+   useEffect(() => setPullFilter('external_block', showBlocked ? null : isPullBlocked), [showBlocked]);
 
    return (
       <Center py={2} bgColor="var(--header-background)" color="var(--brand-color)" {...props}>
--- a/frontend/src/navbar.tsx
+++ b/frontend/src/navbar.tsx
@@ -16,7 +16,6 @@ import {
    MenuList,
    Checkbox,
    CheckboxGroup,
-   Stack,
 } from "@chakra-ui/react";
 import { useRef, useEffect, useCallback } from "react";
 import { useBoolUrlState } from "./use-url-state";
--- a/frontend/src/navbar.tsx
+++ b/frontend/src/navbar.tsx
@@ -40,10 +40,10 @@ export function Navbar(props: BoxProps) {
    const hideBelowLarge = ['none', 'none', 'none', 'block'];
 
    const toggleShowCryo = useCallback(() => setShowCryo(!showCryo), [showCryo]);
-   useEffect(() => setPullFilter('cryo', showCryo ? null : isPullCryo), [showCryo]);
+   useEffect(() => setPullFilter('cryo', showCryo ? null : isNotCryogenic), [showCryo]);
 
    const toggleShowExtBlocked = useCallback(() => setShowExtBlocked(!showExtBlocked), [showExtBlocked]);
-   useEffect(() => setPullFilter('external_block', showExtBlocked ? null : isPullExtBlocked), [showExtBlocked]);
+   useEffect(() => setPullFilter('external_block', showExtBlocked ? null : isNotExternallyBlocked), [showExtBlocked]);
 
    return (
       <Center py={2} bgColor="var(--header-background)" color="var(--brand-color)" {...props}>
@@ -125,11 +125,11 @@ export function Navbar(props: BoxProps) {
    );
 }
 
-function isPullCryo(pull: Pull): boolean {
+function isNotCryogenic(pull: Pull): boolean {
    return !pull.getLabel("Cryogenic Storage");
 }
 
-function isPullExtBlocked(pull: Pull): boolean {
+function isNotExternallyBlocked(pull: Pull): boolean {
    return !pull.getLabel("external_block");
 }
 
--- a/frontend/src/navbar.tsx
+++ b/frontend/src/navbar.tsx
@@ -11,6 +11,12 @@ import {
    Box,
    BoxProps,
    Input,
+   Menu,
+   MenuButton,
+   MenuList,
+   Checkbox,
+   CheckboxGroup,
+   Stack,
 } from "@chakra-ui/react";
 import { useRef, useEffect, useCallback } from "react";
 import { useBoolUrlState } from "./use-url-state";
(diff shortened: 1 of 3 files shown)


rickisXP:
QA :no_entry_sign: 
- Externally blocked filter toggle works as expected :heavy_check_mark: 

deploy_block :thinking: on a consideration
- By default, externally blocked pulls are not shown. I think blocked pulls should be shown by default, and they should be filtered out when the user selects the filter

rickisXP:
QA :+1: carryover and un_deploy_block :+1: on changing initial state

frontend/src/navbar.tsx
```diff
@@ -29,12 +29,16 @@ export function Navbar(props: BoxProps) {
...
    const {toggleColorMode} = useColorMode();
    const [showCryo, setShowCryo] = useBoolUrlState('cryo', false);
+   const [showBlocked, setShowBlocked] = useBoolUrlState('external_block', false);
    const hideBelowMedium = ['none', 'none', 'block'];
    const hideBelowLarge = ['none', 'none', 'none', 'block'];
 
    const toggleShowCryo = useCallback(() => setShowCryo(!showCryo), [showCryo]);
    useEffect(() => setPullFilter('cryo', showCryo ? null : isPullCryo), [showCryo]);
 
+   const toggleShowBlocked = useCallback(() => setShowBlocked(!showBlocked), [showBlocked]);
```
ardelato, line 39: I am not sure about using just `Blocked` as part of these names since we have other things besides `External Blocked` that can be blocked like `CI Blocked`, `Deploy Blocked`, and `Dev Blocked`. 

What about using `ExtBlocked`?

ardelato (approved):
CR 👍 , but deploy_block ☁️ on possibly renaming things related to `external_block`

frontend/src/navbar.tsx
jordycosta: I agree that this could have a better name convention, I like **ExtBlocked** ✅

jordycosta:
QA 🟢 Carry-over
un_deploy_block 🟩 

Updated variable names

ardelato (approved):
CR 👍

sumukhipandey (approved):
CR 🦋 Thanks Jordy

jordycosta:
QA 🟢 Carry-over

Updated the button behavior for a cleaner UI

firefox_lCJV4cUkdk

frontend/src/navbar.tsx
```diff
@@ -63,9 +63,9 @@ export function Navbar(props: BoxProps) {
                   size="sm"
                   title="Toggle display for externally blocked pull requests"
                   colorScheme="blue"
-                  variant={showExtBlocked ? 'solid' : 'ghost'}
+                  variant='ghost'
                   onClick={toggleShowExtBlocked}>
-                  <FontAwesomeIcon icon={faEyeSlash}/>
+                  <FontAwesomeIcon icon={showExtBlocked ? faEyeSlash : faEye}/>
```
ardelato, line 68: What was the reasoning for this change?

From my perspective, I associate the `faEyeSlash` icon with non-QAable pulls and the `faEye` icon with pulls currently being QAed. Therefore, each icon has its own meaning. However, it now sees that the toggling will change the icon used so this may cause some confusion.

frontend/src/navbar.tsx
```diff
@@ -86,6 +99,10 @@ function isPullCryo(pull: Pull): boolean {
    return !pull.getLabel("Cryogenic Storage");
 }
 
+function isPullExtBlocked(pull: Pull): boolean {
+   return !pull.getLabel("external_block");
+}
```
mlahargou, line 104: This logic seems backwards to me... shouldn't this function return true if the `external_block` label is on the PR? Aren't we currently saying "if the pull doesn't have the `external_block` label, return true?"

frontend/src/navbar.tsx
jordycosta: Loading up Pulldasher before this change would have the button selected initially like this

selected

This change was just a quick way to determine if we are showing externally blocked pull requests or not. I didn't really think about the QA icon there since this button icon would initially be shown as the faEyeSlash and only then changed when someone wants to hide externally blocked PRs; I don't really think it could cause much confusion, but I do understand the concern.

jordycosta: I'll keep this as just the faEyeSlash and then have the border appear around it if it's active (like Cryogenic Storage)

frontend/src/navbar.tsx
jordycosta: **getLabel** returns the label string using find(), if whatever passed as a parameter is present on a specific pull

     * Returns the value of the first element in the array where predicate is true, and undefined
     * otherwise.

The `!` here will make this function return a boolean value based on it's return value

mlahargou: Derp, thanks for the info!

jordycosta: https://github.com/iFixit/pulldasher/commit/35ba36c68c1a3b2f134f477307ea570a3e1010d2

frontend/src/navbar.tsx
jordycosta: https://github.com/iFixit/pulldasher/pull/334/commits/24c603dd10354fd8b984c81cc1e72e97c57760dd

ardelato (approved):
CR 🦖

jordycosta:
QA 🟢 Carry-over

Only changed button attributes

firefox_8HQxXq47Iv

sumukhipandey (approved):
CR 🦋

frontend/src/navbar.tsx
danielbeardsley: I think it's just function name that's wrong. If the label is present, `getLabel` will return something truthy and this function will negate that and return false. I think it should be named `isNotExternallyBlocked`. The same goes for `isPullCryo` for that matter.

danielbeardsley:
Looks great!  CR :+1: 

dev_block :+1: on a few function renames though.

danielbeardsley:
Hmm.. Also, I think the eye icon isn't appropriate here cause several other icons *also* hide and show a set of pulls. We want the icon to loosely represent which pulls are hidden / shown. The "external block" term doesn't lend itself to easy translation to an icon though.

Maybe it's time to turn this into a drop-down list like the author and repo filters...

jordycosta:
> Maybe it's time to turn this into a drop-down list like the author and repo filters...

What do you think about a button box instead of using the external_block icon?

firefox_HCfDXrYRfY

danielbeardsley:
> What do you think about a button box instead of using the external_block icon?

That looks quite verbose (takes up a lot of room) and I think the button always showing text that's opposite to the current state is confusing. Maybe a dropdown like the others called `Filters` or something and has `Externally Blocked` and `Cryogenic Storage` entries ?

danielbeardsley:
> Maybe a dropdown like the others called Filters or something and has Externally Blocked and Cryogenic Storage entries 

Hmm..  this may work better if we reverse the terminology and use `Hide Externally Blocked`?

Trying to imagine how this works:
```
X Show All
 - Hide Externally Blocked
 - Hide Frozen
```

Clicking `Hide Frozen` should hide the Frozen pulls and uncheck it, so it then looks like this?

```
 - Show All
 - Hide Externally Blocked
 X Hide Frozen
```
Then clicking "Hide externally blocked" will check that box:

```
 - Show All
 X Hide Externally Blocked
 X Hide Frozen
```

Then clicking "Show All" will go back to the initial state with "Show All" checked.

jordycosta:
@danielbeardsley 

How about something like this?

firefox_rDpe41asQE

It will keep the default state of the two label variables so Cryogenic will be hidden initially like before, and externally block will be shown until toggled off here

danielbeardsley:
> How about something like this?

That's looking good! I'd still opt for using a menu list with checks like the other dropdowns so you can see the state, but that looks great.

jordycosta:
> I'd still opt for using a menu list with checks like the other dropdowns so you can see the state
> @danielbeardsley 

firefox_d5OI5JoAH3

I figured out how to get this working using the [Chakra Menu](https://v1.chakra-ui.com/docs/components/overlay/menu) component

The only roadblock I see now is that the external_block should be checked initially

```
<MenuList>
   <MenuOptionGroup type="checkbox">
      <MenuItemOption isChecked={showCryo} value="cryogenic" onClick={toggleShowCryo} paddingLeft="184px" iconSpacing="-180px">Cryogenic Storage</MenuItemOption>
      <MenuItemOption isChecked={showExtBlocked} value="external_block" onClick={toggleShowExtBlocked} paddingLeft="184px" iconSpacing="-180px">External Block</MenuItemOption>
   </MenuOptionGroup>
 </MenuList>
```

The two ways I'm thinking of resolving this are:

1.) Switch external_block to not be shown initially (this may not be the best idea since CR is still necessary for these pulls)
2.) Figure out how to set the initial state of this MenuOptionGroup to the external_block value

frontend/src/navbar.tsx
jordycosta: Resolved via https://github.com/iFixit/pulldasher/pull/334/commits/f373bca3fabe86ffc85afaea2dd68f460a1bb408

jordycosta:
un_dev_block 🟩 

Updated function names + new drop down with state indicators

danielbeardsley:
Functionally, it works great! Stylistically, it's different enough from the other dropdowns that it seems... off. How do you feel about pulling out some re-usable styled pieces so it looks cohesive?

jordycosta:
> Functionally, it works great! Stylistically, it's different enough from the other dropdowns that it seems... off. How do you feel about pulling out some re-usable styled pieces so it looks cohesive?

I do agree that there are some distinct differences between the other dropdowns, and I am open to adjusting to that styling. 

I'll adjust the button itself to match the looks of the other ones, and I'm looking into getting the hover effect

jordycosta:
un_dev_block 🟩 
Updated styling for the dropdown component

rickisXP:
QA :+1: 
- By default, external_block is initially selected/active :heavy_check_mark: 
- Filters behave as expected :heavy_check_mark:

danielbeardsley:
I tried using `MenuItemOption` so the styling would match the other menus exactly. Looks great and a bit less custom styling!\

CR :+1:

jordycosta:
QA 🟢 Carry-over

> I tried using MenuItemOption so the styling would match the other menus exactly

Thank you!

jarstelfox (approved):
CR 🌵

PR #336, opened 2022-09-19
Reviews: Treat reviews as CR status

mlahargou:
Reviews are awesome in GH. They have a natively defined state in GH which
we can read in pulldasher. Let's pull that information automatically
from the Review.

The old `CR :emoji` should still work on regular comments. We plan to
phase this out if we like working strictly with Reviews for the CR
status.

Note: We now don't listen to any other signatures on the Review comment.

QA:
--
Check out this branch on pulldasher-dev and play around with it. Make sure
`CR :emoji:` get invalidated when a new commit is pushed. Make sure
Reviews are invalidated when the Review is dismissed.

Connects: https://github.com/iFixit/ifixit/issues/44538
Connects: https://github.com/iFixit/ifixit/discussions/44648

cc @danielbeardsley

### DIFF
--- a/models/signature.js
+++ b/models/signature.js
@@ -35,28 +35,11 @@ Signature.parseComment = function parseComment(comment, repoFullName, pullNumber
       id: comment.id
    }
 
-   return parseSignatures(commentData);
+   return parseCommentSignatures(commentData);
 };
 
-/**
- * Parses a GitHub review and returns an array of Signature objects.
- * Returns an empty array if the comment did not contain any of our tags.
- */
-Signature.parseReview = function parseReview(review, repoFullName, pullNumber) {
-   const reviewData = {
-      repo: repoFullName,
-      number: pullNumber,
-      body: review.body,
-      user: review.user,
-      created_at: review.submitted_at,
-      id: review.id
-   }
-
-   return parseSignatures(reviewData);
-};
-
-function parseSignatures(data) {
-	let signatures = [];
+function parseCommentSignatures(data) {
+   let signatures = [];
 
    config.tags.forEach(function(tag) {
       if (hasTag(data.body, tag)) {
@@ -74,6 +57,75 @@ function parseSignatures(data) {
          }));
       }
    });
+
+   return signatures;
+}
+
+/**
+ * Parses a GitHub review and returns an array of Signature objects.
+ * Returns an empty array if the comment did not contain any of our tags.
+ */
+Signature.parseReview = function parseReview(review, repoFullName, pullNumber) {
+   const reviewData = {
+      repo: repoFullName,
+      number: pullNumber,
+      body: review.body,
+      user: review.user,
+      created_at: review.submitted_at,
+      id: review.id,
+      state: review.state
+   }
+
+   return parseReviewSignature(reviewData);
+};
+
+function parseReviewSignature(data) {
+   let signatures = [];
+
+   if (data.state === 'APPROVED') {
+      signatures.push(new Signature({
+         repo: data.repo,
+         number: data.number,
+         user: {
+            id:    getUserid(data.user),
+            login: getLogin(data.user)
+         },
+         type: 'CR',
+         created_at: data.created_at,
+         active: true,
+         comment_id: data.id
+      }));
+   };
+
+   if (data.state === 'DISMISSED' || data.state === 'PENDING') {
+      signatures.push(new Signature({
+         repo: data.repo,
+         number: data.number,
+         user: {
+            id:    getUserid(data.user),
+            login: getLogin(data.user)
+         },
+         type: 'CR',
+         created_at: data.created_at,
+         active: false,
+         comment_id: data.id
+      }));
+   }
+
+   if (data.state === 'CHANGES_REQUESTED') {
+      signatures.push(new Signature({
+         repo: data.repo,
+         number: data.number,
+         user: {
+            id:    getUserid(data.user),
+            login: getLogin(data.user)
+         },
+         type: 'dev_block',
+         created_at: data.created_at,
+         active: true,
+         comment_id: data.id
+      }));
+   }
 
    return signatures;
 }
--- a/models/signature.js
+++ b/models/signature.js
@@ -95,7 +95,7 @@ function parseReviewSignature(data) {
          active: true,
          comment_id: data.id
       }));
-   };
+   }
 
    if (data.state === 'DISMISSED' || data.state === 'PENDING') {
       signatures.push(new Signature({
--- a/models/signature.js
+++ b/models/signature.js
@@ -80,54 +80,31 @@ Signature.parseReview = function parseReview(review, repoFullName, pullNumber) {
 };
 
 function parseReviewSignature(data) {
-   let signatures = [];
+   let signatureData = {
+      repo: data.repo,
+      number: data.number,
+      user: {
+         id:    getUserid(data.user),
+         login: getLogin(data.user)
+      },
+      created_at: data.created_at,
+      comment_id: data.id
+   };
 
    if (data.state === 'APPROVED') {
-      signatures.push(new Signature({
-         repo: data.repo,
-         number: data.number,
-         user: {
-            id:    getUserid(data.user),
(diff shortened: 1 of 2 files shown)


models/signature.js
```diff
@@ -78,6 +61,75 @@ function parseSignatures(data) {
...
+      state: review.state
+   }
+
+   return parseReviewSignature(reviewData);
+};
+
+function parseReviewSignature(data) {
+   let signatures = [];
+
+   if (data.state === 'APPROVED') {
```
danielbeardsley, line 85: Given that all these ifs are mutually exclusive, I'd opt for an `if / else if` here so the code makes that obvious.

models/signature.js
```diff
@@ -78,6 +61,75 @@ function parseSignatures(data) {
...
+         },
+         type: 'CR',
+         created_at: data.created_at,
+         active: true,
+         comment_id: data.id
+      }));
+   }
+
+   if (data.state === 'DISMISSED' || data.state === 'PENDING') {
+      signatures.push(new Signature({
```
danielbeardsley, line 101: I'm tempted to say that we should create the object up top and only mutate the `type` and `active` in the conditions. Not for code reduction, but to reduce the risk of a change updating one of these ifs and forgetting the other.

Thoughts? This is minor.

danielbeardsley (approved):
CR :+1: 

Great job! I'm glad this wasn't that many changes.

I just had a few suggestions about small structure, but I'm fine with it as-is too.

models/signature.js
mlahargou: The problem with this is that there is also a `COMMENT` status. We don't want to do anything on that I don't think. I could just bail early if it's a `COMMENT`.

danielbeardsley: I see.

image

Should we interpret comments as anything? Maybe deploy block? i.e. someone is saying "I've looked at this and I have questions I want addressed before I approve, but I don't have specific changes I want".

I feel like many of my CRs end up in that state. Thoughts?

danielbeardsley:
I know this will be a confusing screenshot... but GH is still saying my review "approves" this pull even though you've pushed new commits. I don't get it, I assumed a new review would be required after making changes.

image

dev_block :+1:  I'm a lot less confident in using reviews is new commits don't invalidate old reviews...

models/signature.js
mlahargou: If the type of the review is not one that we care about (i.e. `COMMENT`), we ignore it now.

mlahargou: Sometimes I just comment `"Nice!"` or something like that on the comment. I think we just ignore it.

mlahargou:
> I know this will be a confusing screenshot... but GH is still saying my review "approves" this pull even though you've pushed new commits. I don't get it, I assumed a new review would be required after making changes.

Yeah, what? That would totally defeat the purpose of this...

mlahargou:
I wonder if the problem is with this setting: 
image

That's a branch protection rule for `master`.

mlahargou:
Derp! It was turned off for this repo!

image

danielbeardsley:
> Derp! It was turned off for this repo! I was looking at ifixit/ifixit before.

Hmm...  didn't realize it could be turned off like that... hmm..  Makes me wonder if we need to validate that setting on boot or something (if it's even available via the API). Is that turned on by default?

mlahargou:
> Hmm... didn't realize it could be turned off like that... hmm.. Makes me wonder if we need to validate that setting on boot or something (if it's even available via the API). Is that turned on by default?

Not sure what you mean by 'validate on boot?' Do you mean on repo creation? I believe it's turned off by default.

danielbeardsley:
> Not sure what you mean by 'validate on boot?

I meant validate the repo settings during pulldasher startup (we already query for all open pulls). If we don't have that setting set, fail to start, or ignore that repo, or .... something. Just an idea.

> I believe it's turned off by default.

Ugh.

danielbeardsley:
Maybe we can compare review dates and commit dates (like we do with comments) and only consider the reviews that came *after* the last comit.... ugh, but then we're back to merges invalidating reviews.... maybe we only do that for repos that don't have the switch on?

mlahargou:
> Maybe we can compare review dates and commit dates (like we do with comments) and only consider the reviews that came after the last comit.... ugh, but then we're back to merges invalidating reviews.... maybe we only do that for repos that don't have the switch on?

I feel like we'd want this turned on for all repos, right? I feel like we'd notice this pretty quickly if we forgot to turn it on.

danielbeardsley:
> I feel like we'd want this turned on for all repos, right? I also feel like we'd notice this pretty quickly if we forgot to turn it on.

Yes, I don't know why we wouldn't want it turned on, when off reviews seem a little pointless.

>  I also feel like we'd notice this pretty quickly if we forgot to turn it on.

I'm less confident, given that we have *many* repos, but still, I don't think it's worth extra code here, so yeah, I don't think we should do anything about it.

danielbeardsley:
Hmm... one more thing that we'll have to tell QA (sorry folks) is that they can no longer use native GH reviews (otherwise they'll be interpreted as CR). Maybe you should merge this and make the announcement.

danielbeardsley (approved):
CR :+1:

erinemay:
> Hmm... one more thing that we'll have to tell QA (sorry folks) is that they can no longer use native GH reviews (otherwise they'll be interpreted as CR). Maybe you should merge this and make the announcement.

I think this should be ok. We're talking about it today. I think the most we do is `comment`, and it looks like we're ignoring those. We shouldn't be (and haven't been) leaving `approve` via GH reviews. I can imagine a few cases in which we might `request changes`. What's the flow for that? Does it count as CR? Or is that equivalent to dev_block?

mlahargou:
> What's the flow for that? Does it count as CR? Or is that equivalent to dev_block?

That would be a dev_block

jordycosta (changes requested):
test 2

jordycosta:
dev_block 🟥 on some potential bugs

Requesting a review on a pull request will place it in a dev_block state
> Dismissing these requested changes will un_dev_block the pull request
> Approving these requested changes **does not** un_dev_block the pull request

---

Also, if a pull request's changes are approved and new commits are pushed after the approval for some reason: should this invalidate that equivalent CR stamp (from the approved review before the additional commits)?

mlahargou:
I found the bug. We dev_block on `CHANGES_REQUESTED`. If those get approved, we CR. That dev_block still remains though.

The only way around this is to put a un_dev_block signature when we get an `APPROVED` Review. I see problems with this though:

- Reviewer1 comes in and requests changes
  - This puts a dev_block on the PR
- Reviewer2 comes in and approves the PR
  - This puts a CR on the PR
  - This puts an un_dev_block on the PR

Now, Reviewer1's dev_block just got dismissed seemingly without reason. I'm not sure what to do here. Do you have any ideas @danielbeardsley.

jordycosta:
@mlahargou 

I tested your scenario on https://github.com/iFixit/ifixit/pull/44736

The second reviewer creating their own approval via a review won't change the dev_block state

Since the second reviewer's approval doesn't affect the first reviewer's requested changes, the dev_block state would only depend on that first request being dismissed or approved after the changes are made

mlahargou:
> The second reviewer creating their own approval via a review won't change the dev_block state

> Since the second reviewer's approval doesn't affect the first reviewer's requested changes, the dev_block state would only depend on that first request being dismissed or approved after the changes are made

I'm saying that **if** I made the proposed changes, the second reviewer's `APPROVED` Review would cancel the dev_block.

jordycosta:
> I'm saying that **if** I made the proposed changes [in my comment](https://github.com/iFixit/pulldasher/pull/336#issuecomment-1255360683), the second reviewer's `APPROVED` Review would cancel the dev_block.

Ah! I see, I didn't think the requested changes approval was the same as a normal review approval

---

Thinking about it a little more, wouldn't the second reviewer only give approval if those requested changes were made?

or

The second reviewer could always include a dev_block with a link to the requested changes comment, but that may require adding in changes to include those signatures in review comments also.

PR #341, opened 2022-11-04
GitHub webhooks: be tolerant

danielbeardsley:
* Gracefully deal with null `pull.body`
* Support parsing larger webhook bodies (we've see pretty big ones)

qa_req 0 - I think there's not much to QA here without actually trying it


andyg0808 (approved):
CR :+1:

PR #350, opened 2023-01-14
Commit Statuses: add started / completed times 

danielbeardsley:
Add code to parse started/completed times when dealing with commit statuses and check runs. Include them all through the code and models, making them available on the front end. Also, add a small "duration" indicator next to each CI status.

Note: The main intent for these times is to be able to make a dashboard for CI, the duration UI was just added to demonstrate the data.

image

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
          DEBUG: pulldasher*         # Enables all debug prints and opens 127.0.0.1:9229 for remote debugging
          MOCK_AUTH_AS_USER: "fakeuser"
       ports:
-         - 8000:3000                # Map the container's port 8080 to the host's 8080
+         - 8000:3000                # Map the container's port 3000 to the host's 8000
                                     # NOTE: The host port will need to match
                                     # the callback URL specified in the GitHub
                                     # application setting.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
          DEBUG: pulldasher*         # Enables all debug prints and opens 127.0.0.1:9229 for remote debugging
          MOCK_AUTH_AS_USER: "fakeuser"
       ports:
-         - 3000:3000                # Map the container's port 8080 to the host's 8080
+         - 8000:3000                # Map the container's port 8080 to the host's 8080
                                     # NOTE: The host port will need to match
                                     # the callback URL specified in the GitHub
                                     # application setting.
--- a/frontend/src/pull-card/commit-statuses.tsx
+++ b/frontend/src/pull-card/commit-statuses.tsx
@@ -57,6 +57,9 @@ function StatusLink({status}: {status: CommitStatus}) {
          <chakra.span>
             {status.data.description}
          </chakra.span>
+         <chakra.span>
+            <Duration status={status}/>
+         </chakra.span>
       </StatusLinkContainer>
    );
 }
@@ -130,3 +133,14 @@ function CommitStatuses({pull}: {pull: Pull}) {
    </Popover>
    );
 });
+
+function Duration({status}: {status:CommitStatus}) {
+   const {started_at, completed_at} = status.data;
+   if (!started_at) {
+      return <>Queued</>;
+   }
+
+   const completed = completed_at || Date.now() / 1000;
+   const min = Math.ceil((completed - started_at)/60);
+   return <>{min}m</>;
+}
--- a/frontend/src/pull-card/commit-statuses.tsx
+++ b/frontend/src/pull-card/commit-statuses.tsx
@@ -1,6 +1,6 @@
 import { Pull } from '../pull';
 import { CommitStatus, StatusState } from '../types';
-import { newTab } from "../utils";
+import { newTab, useDurationMinutes } from "../utils";
 import {
    chakra,
    Box,
@@ -43,6 +43,7 @@ const StatusLinkContainer = Box;
 function StatusLink({status}: {status: CommitStatus}) {
    const styles = useStyleConfig('StatusLink', {variant: status.data.state});
    const link = status.data.target_url;
+   const duration = useDurationMinutes(status);
    return (
       <StatusLinkContainer
          __css={styles}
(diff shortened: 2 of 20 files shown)


deltuh-vee:
QA 🎬
Pulls now include CI runtimes

Which CI check is this?
image

danielbeardsley:
> Which CI check is this?

I remember seeing that, but I don't know where.

docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
          DEBUG: pulldasher*         # Enables all debug prints and opens 127.0.0.1:9229 for remote debugging
          MOCK_AUTH_AS_USER: "fakeuser"
       ports:
-         - 3000:3000                # Map the container's port 8080 to the host's 8080
+         - 8000:3000                # Map the container's port 8080 to the host's 8080
```
andyg0808, line 11:
```suggestion
         - 8000:3000                # Map the container's port 3000 to the host's 8000
```

frontend/src/pull-card/commit-statuses.tsx
```diff
@@ -130,3 +133,14 @@ function CommitStatuses({pull}: {pull: Pull}) {
...
+
+function Duration({status}: {status:CommitStatus}) {
+   const {started_at, completed_at} = status.data;
+   if (!started_at) {
+      return <>Queued</>;
+   }
+
+   const completed = completed_at || Date.now() / 1000;
+   const min = Math.ceil((completed - started_at)/60);
+   return min ? <>{min}m</> : null;
```
andyg0808, line 145: Might consider moving the `chakra.span` into here, so we can skip rendering the `<span>` if it's going to be empty.

andyg0808 (commented):
CR :+1: 
deploy_block :+1: on a few minor comments.

models/db_status.js
```diff
@@ -2,7 +2,7 @@ var _ = require('underscore'),
     db = require('../lib/db');
 
 // Builds an object representation of a row in the DB `commit_statuses`
-// table from the data returned by GitHub's API.
+// table from an instance of the Status model
```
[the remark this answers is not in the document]
danielbeardsley: Oh I agree... if this were typescript. Alas... the backend is not processed with typescript at all. Just `frontend/*`.

frontend/src/pull-card/commit-statuses.tsx
danielbeardsley: Fixed.

docker-compose.yml
danielbeardsley: Fixed.

lib/db-manager.js
```diff
@@ -151,6 +154,8 @@ var dbManager = module.exports = {
          description:   desc,
          target_url:    url,
          context:       context,
+         started_at:    body.check_run.started_at,
```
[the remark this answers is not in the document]
danielbeardsley: This is outdated and has been moved to the `githubhooks` module. Note: check_runs are different than statuses (and have a different webhook body).

docker-compose.yml
danielbeardsley: Fixed.

k0rvusk0r4x:
I'm not seeing the times now. 
image
I used [this guide](https://overlord.ifixit.com/Guide/How+to+test+Pulldasher+with+Pulldasher-dev/1124) to test. Is it no longer accurate?

dev_block 😿

frontend/src/utils.ts
```diff
@@ -23,11 +23,54 @@ export function userProfileUrl(username: string): string {
...
 }
 
 export function newTab(url: string) {
    window.open(url, "_blank");
 }
+
+function fakeCIStatuses(pulls: PullData[]) {
+   const fakeContexts = ['psalm', 'deploy', 'unit-tests', 'api', 'integration', 'browser', 'bundle-analysis'];
+   pulls.forEach((pull) => {
+      pull.status.commit_statuses = fakeContexts.map((context) => {
```
andyg0808, line 38: This looks like it's mutating the original objects in `pulls` (which might be confusing down the road, because we're also returning the mutated objects). If we don't want to mutate, we could consider using something like this:
```ts
return pulls.map((pull) => {
   const commit_statuses = fakeContexts.map((context) => {
       const now = unix();
       const startedAt = now - (Math.random() * 120 * 60);
       const duration = Math.random() * 10 * 60;
       const completedAt = startedAt + duration;
       return {data: {
          sha: "fake sha",
          target_url: "https://example.com",
          description: context,
          context: context,
          state: completedAt > now ? StatusState.pending : finishedState(),
          started_at: startedAt,
          completed_at: completedAt > now ? null : completedAt,
       }};
   });
   return {
      ...pull,
      status: {
         ...pull.status,
         commit_statuses,
      },
   }
```

frontend/src/utils.ts
```diff
@@ -23,11 +23,54 @@ export function userProfileUrl(username: string): string {
...
+         }};
+      });
+   });
+   return pulls;
+}
+
+function finishedState() {
+   const r = Math.random();
+   return r < 0.01 ? StatusState.error :
+      (r < 0.03 ? StatusState.failure : StatusState.success);
```
andyg0808, line 60: Just to be clear, I think this is
1% of the time, return `StatusState.error`
2% of the time, return `StatusState.failure`
97% of the time, return `StatusState.success`

I'm suspicious that the 2% there is supposed to be 3%; I believe it's only 2% because
`0 <= r < 0.01` => `StatusState.error`
and thus
`0.01 <= r < 0.03` => `StatusState.failure`, which is `0.03-0.01 = 0.02 = 2%`.

andyg0808 (approved):
CR :+1: 
A couple comments, in case you're interested.

frontend/src/utils.ts
danielbeardsley: This is only used on an array of dummy pulls and that's only used in development, I'm OK with mutating.

frontend/src/utils.ts
danielbeardsley: Yes, you are likely right that I was moving fast and didn't think about the 3 - 1 == 2, but I'm OK with that. This isn't critical, it's just to give a sprinkling of different states.

danielbeardsley:
QA :+1: 

un_dev_block :+1: 

I'll save you the time. I'm seeing them:
image

The dev instance of pulldasher is often out of date cause it isn't always running. These values sometimes aren't present until you click "Refresh" on the pull (if it's a pull that missed receiving webhooks when this branch wasn't deployed).

danielbeardsley:
un_deploy_block :+1:

PR #353, opened 2023-02-22
Notifications: Make sound work

andyg0808:
This uses the correct URL for the bell sound which is supposed to be available for Pulldasher.

## QA Notes
1. Run the UI with `npm run start`.
2. Listen for the bell sound when the page loads.

PR #354, opened 2023-02-22
Notification: Open pull on click

andyg0808:
This adds a new feature to the Pulldasher notifications: they now open the pull they're about when you click them. This is much more convenient, since the majority of the time you want to open the pull if you're interacting with the notification.

## QA Notes
1. Run `npm run frontend:start`
2. Click one of the notifications which appears and confirm that the pull is opened in a new tab.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,57 +1,52 @@
 {
-    "extends": [
-       "eslint:recommended"
-    ],
-    "parserOptions": {
-       "ecmaVersion": 2018,
-       "sourceType": "module"
-    },
-    "env": {
-       "browser": true,
-       "node": true,
-       "es6": true
-    },
-    "globals": {
-       "App": true
-    },
-    "ignorePatterns": [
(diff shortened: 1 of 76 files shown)


andyg0808:
deploy_block :+1: on https://github.com/iFixit/pulldasher/pull/353

andyg0808:
dev_block :+1: 
This worksish, but it's awkward to use, because it doesn't manage the focus well. I'd like to see if I can do better.

andyg0808:
I think the ultimate solution to this might be service workers and https://developer.mozilla.org/en-US/docs/Web/API/Clients/openWindow.

sctice-ifixit:
Should this be closed @andyg0808?

andyg0808:
@sctice-ifixit I'd like to do it sometime, but I realize that time may be pretty far off. It's up to you.

davidrans:
Let's close it for now, we can always reopen

PR #368, opened 2023-03-30
Move ready, no-CI pulls

### LINKED ISSUE #164: Move ready no-ci builds into the ready column but keep the no-ci indicator
It's confusing to have them as "Deploy blocked" when they're really "Deploy carefully".

Screen Shot 2019-11-21 at 3 07 27 PM

jordycosta:
**Before/After**

B4
After

qa_req 0
Closes #164

### DIFF
--- a/frontend/src/pull.ts
+++ b/frontend/src/pull.ts
@@ -115,7 +115,7 @@ export class Pull extends PullData {
     return (
       this.hasMetDeployRequirements() &&
       !this.getDevBlock() &&
-      (!!this.getDeployBlock() || !this.isCiRequired() || this.hasMergeConflicts() || this.isDependent())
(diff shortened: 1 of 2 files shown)


danielbeardsley:
Let's ask some opinions here. Issues that have sat around for four years aren't necessarily still relevant.

I think it's actually not CI that's relevant to me here, but whether any other tasks need to happen during deployment. i.e. is clicking "merge" enough, or is there more to do? No-CI is somewhat of a proxy for "deploy is not automatic" and so this kinda serves this purpose.

That said, I don't feel that strongly.

@sterlinghirsh, @davidrans can we get a quick gut reaction?

sterlinghirsh:
I think putting it in ready is fine, especially since it has the alert icon.

PR #373, opened 2023-04-11
Commit Checks: Use new api location

danielbeardsley:
The old checks endpoint isn't supported by github's new Fine Grained
Personal Access Tokens; but this equivalent endpoint is.


danielbeardsley:
QA :+1: 

Doesn't crash anymore when using a FAT!

danielbeardsley:
dev_block :+1: 

No longer crashes, the but API results are a different shape and it seems to be considering them all "not started". Pause.

danielbeardsley:
un_dev_block :+1: 

Had to combine two apis... resulting in six API requests where there used to be just one.

danielbeardsley:
QA :+1: 

It's running and refreshing pulls correctly.

andyg0808 (approved):
CR :+1: 
Looks reasonable to me.

PR #377, opened 2023-04-26
Search: Add repository pattern

### LINKED ISSUE #335: Search: search the repo name as well
We should include the repo name in the search

Specifics:
* A search for "{repo name} {part of the title}" should still find the pull instead of mandating the whole search query is found in the repo name

jordycosta:
### Preview:

_Before_

before

**After**

after

---

Closes #335

### DIFF
--- a/frontend/src/navbar.tsx
+++ b/frontend/src/navbar.tsx
@@ -200,9 +200,7 @@ function SearchInput() {
         "search",
         patterns.length
           ? (pull: Pull) => {
-              const titleMatch = patterns.every((pattern) => pull.title.match(pattern));
-              const repoNameMatch = patterns.every((pattern) => pull.repo.match(pattern));
-              return titleMatch || repoNameMatch;
+              return patterns.every((pattern) => pull.title.match(pattern) || pull.repo.match(pattern));
             }
           : null
       );
--- a/frontend/src/navbar.tsx
+++ b/frontend/src/navbar.tsx
@@ -200,7 +200,9 @@ function SearchInput() {
         "search",
         patterns.length
           ? (pull: Pull) => {
-              return patterns.every((pattern) => pull.title.match(pattern));
+              const titleMatch = patterns.every((pattern) => pull.title.match(pattern));
+              const repoNameMatch = patterns.every((pattern) => pull.repo.match(pattern));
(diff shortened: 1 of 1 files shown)


deltuh-vee:
QA 🎬
Repository names are now also included in search results.

frontend/src/navbar.tsx
```diff
@@ -200,7 +200,9 @@ function SearchInput() {
         "search",
         patterns.length
           ? (pull: Pull) => {
-              return patterns.every((pattern) => pull.title.match(pattern));
+              const titleMatch = patterns.every((pattern) => pull.title.match(pattern));
+              const repoNameMatch = patterns.every((pattern) => pull.repo.match(pattern));
+              return titleMatch || repoNameMatch;
```
danielbeardsley, line 205: I love it. Though now that I think about it, I think it would be more useable to do:

`patterns.every(pattern => title.match(patter) || repo.match(pattern))`

Because, as-is, this requires that your search query has to be entirely found in the repo name, or entirely found in the title.

Most searches are "search for this text anywhere in the searchable fields" i.e. operating as if the searchable fields were all concatenated together instead of treating each field as a separate document.

danielbeardsley (changes requested):
dev_block :+1:  on considering the above comment.

Otherwise, this is a good move.

frontend/src/navbar.tsx
jordycosta: While we are at this, would it be beneficial at all to include the pull number?

jordycosta: Screenshot 2023-04-27 092530

danielbeardsley: I'm not sure I would use it, but it seems harmless.

jordycosta:
un_dev_block 🟢 

Suggestion implemented via https://github.com/iFixit/pulldasher/pull/377/commits/8a6276b9a4c3cc5d4ff28510addfafb1d9579701

danielbeardsley:
CR :+1:

danielbeardsley (approved):
CR :+1:

deltuh-vee:
Carry Over QA 🎬

PR #378, opened 2023-05-01
Fix 'undefined' title

jordycosta:
### Issue:

firefox_N1gRqWmhRi

This pull request will use an empty string if the `title` prop is _falsy_ instead of defaulting to 'undefined'


danielbeardsley:
CR :+1:

danielbeardsley (approved):
CR :+1:

PR #380, opened 2023-05-19
Pull model: don't send status.CR/QA

danielbeardsley:
We only end up using the superset allCR and allQA properties, let's not send these across the wire. We didn't even include them in the typescript types, so we can be sure we're not using them.


danielcliu-ifixit:
CR :v:

jordycosta:
QA 🟢 

Everything is still working as expected 👍 
https://pulldasher-dev.cominor.com/

danielcliu-ifixit (approved):

PR #387, opened 2023-08-24
Add "draft" option

andyg0808:
This adds an option to hide draft pulls on Pulldasher

I tested it with the dummy pulls, and it looks like it does what I want.

### DIFF
--- a/frontend/dummy-pulls.json
+++ b/frontend/dummy-pulls.json
@@ -6414,6 +6414,96 @@
   },
   {
     "repo": "iFixit/ifixit",
+    "number": 45587,
+    "state": "open",
+    "title": "[DRAFT] Newsletter: create promo page",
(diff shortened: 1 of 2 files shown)


ianrohde:
Does this require a `draft` label in github to work? I looked into this briefly earlier, but moved on to other things because I thought that was the case.

andyg0808:
@ianrohde Nope! It uses the `isNotDraft` function to determine whether a pull should be included or not, and that just checks `isDraft()` on the pull:
https://github.com/iFixit/pulldasher/pull/387/files#diff-4f876b47a482dd69d9bcae2cf9f64a78546c47b753d4228e61931a2dc4fa9e31R205-R207

ianrohde (approved):
CR 💯

Issue #393, opened 2023-10-17
Draft pulls can end up in the Ready column


andyg0808:
I'm assuming there's a missing filter here.

![image](https://github.com/iFixit/pulldasher/assets/1283490/66268528-cd68-4ad1-ba61-8e635167a4f2)

PR #394, opened 2023-10-17
Remove drafts from Ready column

jordycosta:
This will filter out draft pull requests out from the **Ready** column

Closes https://github.com/iFixit/pulldasher/issues/393

### DIFF
--- a/frontend/src/pull.ts
+++ b/frontend/src/pull.ts
@@ -123,6 +123,7 @@ export class Pull extends PullData {
   isReady(): boolean {
     return (
       this.hasMetDeployRequirements() &&
+      !this.isDraft() &&
       !this.getDevBlock() &&
       !this.getDeployBlock() &&
       !this.hasMergeConflicts()


davidrans (approved):
CR :coffee:

frontend/src/pull.ts
```diff
@@ -123,6 +123,7 @@ export class Pull extends PullData {
   isReady(): boolean {
     return (
       this.hasMetDeployRequirements() &&
+      !this.isDraft() &&
```
mlahargou, line 126: Maybe this should be in the `hasMetDeployRequirements` function?

jordycosta: We consider a draft as a type of dev-block so I think it would be best grouped with the other conditions that check for blocks

```
      !this.isDraft() &&
      !this.getDevBlock() &&
      !this.getDeployBlock() &&
```

jordycosta: It may be better to group those into their own hasBlock() function

andyg0808:
CR :+1: but I wouldn't oppose another PR to clean up the functions some way or other.

jordycosta:
@kthaler I resolved the external block if you'd like to test this one out

deltuh-vee:
QA 🎬 
Draft pulls are no longer showing up in the Ready column in pulldasher.

PR #402, opened 2024-02-16
Feat: Add ignored statuses

mlahargou:
We previously could only set `requiredStatuses`. This is great if all ci checks run
on every PR. We want to conditionally run different workflows on different PRs, so
we can't make certain statuses requried.

Instead, let's create a new `ignoredStatuses` that the user can set so that pulldasher
ignores those even if they are failing. 

QA:
--
We should verify that ignored statuses don't affect the ci blocked PRs. You can modify
`/mnt/ebs/etc/pulldasher-dev.config.js` and update pulldasher dev to make sure the
correct pulls are in (or out) of the CI Blocked column on pulldasher-dev.cominor.com.

Connects: https://github.com/iFixit/ifixit/issues/52034

cc @danielbeardsley

### DIFF
--- a/config.example.js
+++ b/config.example.js
@@ -74,6 +74,7 @@ module.exports = {
     {
       name: "owner/otherRepo",
       requiredStatuses: ["tests", "build", "codeClimate"],
+      ignoredStatuses: ["coverage"],
     },
   ],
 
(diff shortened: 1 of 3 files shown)


danielbeardsley (approved):
CR :+1: 

Nice!

deltuh-vee:
So according to this configuration,
CleanShot 2024-02-16 at 11 56 15
should we not show the pass fail status of ifixit-nextjs-dev?
CleanShot 2024-02-16 at 11 56 49

deltuh-vee:
QA 🎬 
PRs that are failing with checks in ignoredStatuses are no longer CI blocked.

PR #411, opened 2024-08-30
Replace mysql npm module with mysql2

danielbeardsley:
Our DB recently changed which auth mechanisms were allowed. We disabled the native password option (the only one that the mysql npm module supports). Thankfully, the mysql2 module is way more popular now and supports the auth we want and the api is nearly identical.


sctice-ifixit (approved):
CR :flags:

danielbeardsley:
QA :+1: The dev deployment is working!

PR #413, opened 2024-09-03
DB: fix DB connection and query format 

danielbeardsley:
Pulldasher was totally broken and would boot until it ran a `REPLACE INTO` query and then crash.

oof, see the second change.

First: The .query().then() was just changed to make use of the new
promise interface instead of promisifying the old callback interface.
It wasn't the problem.

Second: Add support for utf8mb3 to the library (we'll open a pull to add
it upstream). See the code comment.

### DIFF
--- a/lib/db-manager.js
+++ b/lib/db-manager.js
@@ -558,7 +558,7 @@ var dbManager = (module.exports = {
    */
   deleteLabels: function deleteLabels(repo, number) {
     debug("Calling deleteLabels for pull #%s in repo %s", number, repo);
-    var q_delete = "DELETE FROM pull_labels WHERE number = ? " + "AND repo = ?";
+    var q_delete = "DELETE FROM pull_labels WHERE number = ? AND repo = ?";
     return db.query(q_delete, [number, repo]);
   },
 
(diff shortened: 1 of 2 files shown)


danielbeardsley:
This seems to be working fine now!

QA :+1: 

I've opened a pull upstream: https://github.com/sidorares/node-mysql2/pull/3007

djmetzle:
Ref: https://github.com/iFixit/server-templates/issues/4731

Are we still using mb3 for the `metrics` schema?

davidrans:
The tables and columbs are all mb4, but some of the defaults are still mb3:
```
mysql> SHOW VARIABLES LIKE 'character_set_server';
SHOW VARIABLES LIKE 'collation_server';
+----------------------+---------+
| Variable_name        | Value   |
+----------------------+---------+
| character_set_server | utf8mb3 |
+----------------------+---------+
1 row in set (0.00 sec)

+------------------+--------------------+
| Variable_name    | Value              |
+------------------+--------------------+
| collation_server | utf8mb3_unicode_ci |
+------------------+--------------------+
1 row in set (0.00 sec)
```

PR #418, opened 2025-01-15
FilterMenu: support a "defaultExcludedValues" prop

### LINKED ISSUE #414: Repos: allow config to exclude certain repos
### Problem
Using an Org-level webhook is really convenient when you have lots of repos. However we have no way to hide certain repos.

### Do This
Add a config option to hide certain repos (maybe an array of strings or regexes?) by default. Effectively, make the "repos" dropdown filter default to having some repos unselected.

CC @sterlinghirsh

danielbeardsley:
It took a long time to get this to behave intuitively. In short, it took
a long to spec it out. How to have it behave when you click one of the
values, how to represent the default state, ...

The feature as implemented:

* FilterMenu has a new prop defaultExcludedValues: string[]
* Without defaultExcludedValues set, it operates exactly as it does now
* When set, the UI has one new "showDefault" state in addition to
  existing "showAll"
* In "showDefault", Pulls matching one of the excluded values are hidden
  (these excluded values appear de-selected in the UI)
  * One click "selects" them and leaves "showDefault" mode
* "Show All" overrides the "default exclusions" and shows everything

Closes #414

### DIFF
--- a/frontend/src/filter-menu.tsx
+++ b/frontend/src/filter-menu.tsx
@@ -133,9 +133,7 @@ function sortValues(values: string[]): string[] {
 }
 
 function arrayDiff<T>(a: T[], b: T[]): T[] {
-  const ret = a.filter((x) => !b.includes(x));
-   console.log(ret);
-   return ret;
+  return a.filter((x) => !b.includes(x));
 }
 
 function empty<T>(array: T[] | undefined): boolean {
--- a/frontend/src/filter-menu.tsx
+++ b/frontend/src/filter-menu.tsx
@@ -69,7 +69,7 @@ export function FilterMenu({
         : (showDefault ? (pull) => !defaultExculdedValues?.includes(extractValueFromPull(pull))
           : (pull) => selectedValuesSet.has(extractValueFromPull(pull)))
     );
-  }, [selectedValues]);
+  }, [selectedValues, defaultExculdedValues]);
 
   return (
     <Menu closeOnSelect={false}>
@@ -92,7 +92,7 @@ export function FilterMenu({
               Show All
             </MenuItemOption>
             <MenuItemOption
-              key="Show All"
+              key="Show Default"
               onClick={() => setSelectedValues([])}
               isChecked={showDefault}
             >
--- a/frontend/src/filter-menu.tsx
+++ b/frontend/src/filter-menu.tsx
@@ -72,6 +72,8 @@ export function FilterMenu({
     );
   }, [selectedValues, defaultExculdedValues]);
 
+  const numberText = showDefault ? "" : (showAll ? allValues.length : selectedValues.length);
+
   return (
     <Menu closeOnSelect={false}>
       <MenuButton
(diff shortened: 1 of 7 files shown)


danielbeardsley:
CC @sterlinghirsh

frontend/src/filter-menu.tsx
```diff
@@ -98,3 +131,17 @@ function sortValues(values: string[]): string[] {
     a.localeCompare(b, undefined, { sensitivity: "base" })
   );
 }
+
+function arrayDiff<T>(a: T[], b: T[]): T[] {
+  const ret = a.filter((x) => !b.includes(x));
+   console.log(ret);
```
jarstelfox, line 137:
```suggestion
```

jarstelfox (approved):
CR 🌵

frontend/src/filter-menu.tsx
```diff
@@ -98,3 +131,17 @@ function sortValues(values: string[]): string[] {
...
+   return ret;
```
danielbeardsley, line 138:
```suggestion
  return a.filter((x) => !b.includes(x));
```

jarstelfox (approved):
CR 🌵

sterlinghirsh:
Glad to see this happening. I think ifixit-schooner-* and PD-Test should probably be added to the default exclusion list.

It used to be that clicking a single repo would show just that repo, which is usually what I want from a menu like that, but now since they all start checked, clicking a single repo hides it. That is reasonably intuitive when they start checked, though there's no way to select none, so if I do just want to see a single repo, it's like 15 clicks. I think if there was a select none option, that would be sufficient. Datadog handles this with a little button on each menu option that says "only", but that might be overkill for this. Maybe also if it could remember my selection in localstorage or something, that would be good.

I think that Select All, Default, and None should be buttons (actions) rather that checkboxes.

In any case I think this is a great step in the right direction, so thanks for working on it.

danielbeardsley:
> Maybe also if it could remember my selection in localstorage or something, that would be good.

All selections are stored in the url. Bookmarking works well and allows several different configurations.

> I think that Select All, Default, and None should be buttons (actions) rather that checkboxes.

That's not a bad idea. I'll try that out.

danielbeardsley:
> I think that Select All, Default, and None should be buttons (actions) rather that checkboxes.

Done! And I've added an "only" instead of "None"

kthaler:
QA 💎 - We can now exclude repos from pulldasher by default. We can still view the excluded repo with the Repo dropdown by clicking Show All, or by clicking the repo name. 

- [x] We can view only 1 repo by clicking the small `only` button while hovering a dropdown item
  - This has been applied to the Author dropdown as well
- [x] `Show Default` shows all of the repos that aren't  

very small deploy_block 🦖 - When you click Show All, the dropdown indicates that only one repo has been selected, instead of all repos. Then if you unselect a repo, and reselect it, the dropdown now indicates that 17 repos have been selected.

Maybe when Show All is selected, we can just show the number of repos that are selected (right now it's 17), and when Show Default, we can hide the number like we currently do?

kthaler:
I think we should keep it simple, and do `Repo (17)` when you click "Select All" and `Repo` when you click `Select Default`. I don't think most devs will ever use "Select All". 

I'm open to `Repo (All)` as well with "Select All", but don't think that's really necessary.

jarstelfox (dismissed):
CR 🌵

danielbeardsley:
> I think we should keep it simple, and do Repo (17) when you click "Select All" and Repo when you click Select Default

Done!

image

Issue #456, opened 2026-06-09
Issues: Track author, closed reason, and assignment date


sctice-ifixit:
Pulldasher records a fair amount of issue metadata in the `issues` table (title, assignee, status, milestone, created/closed dates) but not a few things I'd like for reporting on GitHub stats: who opened the issue, why it was closed, and when it was assigned to its current assignee.

- **Author** isn't stored at all. The `pulls` table already keeps the PR author as `owner`, but issues have no equivalent.
- **Closed reason** isn't stored. GitHub distinguishes `completed` from `not_planned` (the "won't do" case) via the issue's `state_reason`, and that's useful signal we're dropping.
- **Date assigned** isn't stored. We have the current `assignee` but not when they were assigned, so we can't measure how long issues sit before someone picks them up.

## Do This

Capture these on issue ingest (webhook + bulk refresh) and backfill history:

- [ ] `author` from the issue's GitHub `user.login`
- [ ] `state_reason` (`completed` / `not_planned` / `reopened` / null)
- [ ] `date_assigned`, the time the current assignee was assigned

## Background

`author` and `state_reason` are plain fields on the GitHub issue object, so they flow straight through `Issue.getFromGH`. `date_assigned` is trickier: the issue object carries no assignment timestamp, so we have to read it from the issue's events (the most recent `assigned` event for the current assignee). Pulldasher already fetches issue events for label history during a full refresh, so it's available there.

One consequence: today most issue webhooks take a lightweight path that upserts the webhook body directly and never looks at events. To get `date_assigned` (and to keep label attribution correct), I think we should route issue webhooks through the same full refresh that `opened` already uses. That's an extra events API call per issue webhook, which seems fine at our issue volume.

The schema change is three columns on `issues`. Following the pattern from the `mergeable` column (pulldasher [#356](https://github.com/iFixit/pulldasher/pull/356) plus the metrics migration in ifixit [#46835](https://github.com/iFixit/ifixit/pull/46835)), the pulldasher repo carries the migration and `schema.sql` update for posterity, and the actual `ALTER` runs against the `metrics` database through an ifixit migration.

danielbeardsley:
> That's an extra events API call per issue webhook, which seems fine at our issue volume.

Sure.

LGTM!  I'm surprised that we didn't already have issue author.

PR #356, opened 2023-03-08
Pulldasher: Add the 'mergeable' field

jordycosta:
### Issue overview:

1. Pull requests with merge conflicts can end up in the **Ready** column in Pulldasher
2. QA is unable to pull up the latest changes to a pull request if there are conflicts present

example

### Resolution:

* Send pull requests that have fulfilled CR/QA requirements _(but have a merge conflict)_, to the **Deploy Blocked** column.
* Update QA column sorting to place pull requests without merge conflicts above those with merge conflicts.
* Add a new Pulldasher flag to easily discern if a pull request has merge conflicts.

---

### Changes Made:

1. Added in a new field from GitHub _(mergeable)_
2. Create a new flag for the new field
3. Updated Pulldasher **QA** column sorting logic 
4. Updated Pulldasher **Ready** + **Deploy Blocked** column pull logic

>Documentation reference:
https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#get-a-pull-request

---

### CR/QA Notes:

CR: Review to https://github.com/iFixit/pulldasher/pull/356/commits/88db70109f0aa106040bd2afe84d455b2bd26df4

QA:

1. Checkout | merge with [`add-mergeable-field`](https://github.com/iFixit/pulldasher/tree/add-mergeable-field)
2. `update-pulldasher-dev --run .`
3. Navigate to https://pulldasher-dev.cominor.com/
4. Ensure that pull requests with merge conflicts:

    a.) Have a flag applied
    b.) Are in the **Deploy Blocked** column if requirements are fulfilled
    c.) Are sorted lower in the **QA** column

---

Closes #351

PR #457, opened 2026-06-09
Issues: Track author, closed reason, and assignment date

sctice-ifixit:
Connects to #456

We record a fair amount of issue metadata but not who opened an issue, why it was closed, or when it was assigned. This adds three columns to the `issues` table and populates them on ingest, so we can report on those for GitHub stats.

## Changes

- **`author`** (`varchar(255) NOT NULL DEFAULT 'ghost'`) from the issue's GitHub `user.login`, via the existing `getLogin` helper (so deleted users land as `ghost`, same as pulls).
- **`state_reason`** (`varchar(30)`) straight off the issue object (`completed` / `not_planned` / `reopened` / null).
- **`date_assigned`** (`int`, unix time) for the current assignee. This one isn't a field on the issue object, so derive it from the most recent `assigned` event for the assignee.

To get `date_assigned` on every write path, `handleIssueEvent` now routes all issue webhooks through the full refresh (`refreshPullOrIssue` → `parseIssue`), which already fetches issue events. Before, most issue actions took a lightweight path that upserted the webhook body directly and never looked at events. The tradeoff is one extra events API call per issue webhook, which should be fine at our issue volume (it's what `opened` already did). A side effect: this drops the old `getFromGH`-without-labels path and its `reprocessLabels` repair step, since the full refresh derives labels (with attribution) directly.

The `migrations/0019-*.sql` and `schema.sql` changes here are for posterity, following the `mergeable` precedent (#356). The actual `ALTER` on the `metrics` database ships as a separate ifixit migration ([#62493](https://github.com/iFixit/ifixit/pull/62493), following the [#46835](https://github.com/iFixit/ifixit/pull/46835) precedent). This PR stays `deploy_block`ed until that migration lands, since the new code writes the columns and would error against an un-migrated table.

After both deploy, a `bin/refresh-all-issues` run backfills the three fields across history.

## CR/QA

Tested against pulldasher-dev (see the [Overlord guide](https://overlord.ifixit.com/Guide/How+to+test+Pulldasher+with+Pulldasher-dev/1124)):

- [x] Check out / merge this branch and `update-pulldasher-dev --run .`
- [x] Open https://pulldasher-dev.cominor.com/ and confirm it starts cleanly
- [x] Open, assign, and close an issue in a tracked repo, then confirm the `issues` row has `author`, `state_reason`, and `date_assigned` populated
- [x] Exercise the other issue events that now run through the full refresh and confirm the row stays correct:
  - [x] Reopen the issue: `state_reason` becomes `reopened`
  - [x] Reassign to someone else: `date_assigned` updates to the new assignment time
  - [x] Unassign: `date_assigned` clears
  - [x] Edit the title and add/remove a label: the row and its labels (including `difficulty`) stay correct


sctice-ifixit:
deploy_block :seedling: on running the `ifixit` migration.

models/issue.js
```diff
@@ -72,7 +72,10 @@ class Issue {
       repo: data.repo,
       number: data.number,
       title: data.title,
+      author: getLogin(data.user),
       status: data.state,
+      state_reason: data.state_reason,
+      date_assigned: utils.fromDateString(data.date_assigned),
       date_created: new Date(data.created_at),
```
danielbeardsley, line 79: Hmm... why is this one date column different?

danielbeardsley:
CR :+1:

mlahargou (approved):
CR 📱

models/issue.js
sctice-ifixit: Because `date_created` is required and always a date string, and [`utils.fromDateString(…)`](https://github.com/iFixit/pulldasher/blob/1118c5a4155eb579b27a0a190f7670f90830403b/lib/utils.js#L43) just checks for a `Date` or a falsey value and otherwise does `new Date(…)`. The other date fields here [_can_ be null](https://docs.github.com/en/rest/issues/issues?apiVersion=2026-03-10#get-an-issue). It could just as well be `fromDateString(…)`, but I think it's safe as `new Date(…)` and a little more precise.

sctice-ifixit:
QA :tulip: per the description.

sctice-ifixit:
un_deploy_block :tulip: Migration merged and applied

PR #463, opened 2026-06-16
Honor GitHub PR approvals as CR signoffs

ianrohde:
## Summary
- Treat a GitHub native **Approve** review (`state: APPROVED`) as a CR signoff, so reviewers who approve on GitHub no longer need to add a `CR :emoji:` comment.
- Gated behind `useGithubApprovalForCr` (default on); skipped when the review body already produces a CR signature.

## Problem
Pulldasher had no concept of a GitHub PR approval. CR signoff required manually adding a `CR :emoji:` comment on every PR, even when the reviewer had already approved natively on GitHub.

## Solution
- `models/signature.js`: `parseReview` maps an `APPROVED` review to a CR signature, deduped against an existing `CR` body tag. Works for both the `pull_request_review` webhook and full PR refresh.
- This is read-only: Pulldasher reflects approvals made on GitHub. It does not submit approvals on a user's behalf.

## Deployment notes
- Enable the **Pull request reviews** webhook event on tracked repositories.
- Set `useGithubApprovalForCr: false` to keep legacy emoji-only behavior.

## Test plan
- [x] `npm test` (APPROVED-to-CR mapping, dedup, CHANGES_REQUESTED ignored)
- [x] `npm run lint`
- [ ] Manual: approve a PR on GitHub, confirm the CR bubble fills in Pulldasher
- [ ] Manual: dismiss the approval on GitHub, confirm the CR signoff drops after refresh

### DIFF
--- a/models/signature.js
+++ b/models/signature.js
@@ -54,9 +54,32 @@ Signature.parseReview = function parseReview(review, repoFullName, pullNumber) {
     user: review.user,
     created_at: review.submitted_at,
     id: review.id,
+    state: review.state,
   };
 
-  return parseSignatures(reviewData);
+  const signatures = parseSignatures(reviewData);
+
+  // A GitHub "Approve" review counts as a CR signoff even without a
+  // `CR :emoji:` tag in the body. Skip if the body already produced one.
+  if (
+    useGithubApprovalForCr() &&
+    review.state === "APPROVED" &&
+    !signatures.some((sig) => sig.data.type === "CR")
+  ) {
+    signatures.push(
+      new Signature({
+        repo: repoFullName,
+        number: pullNumber,
+        user: review.user,
+        type: "CR",
+        created_at: review.submitted_at,
+        active: true,
+        comment_id: review.id,
+      })
+    );
+  }
+
+  return signatures;
 };
 
 function parseSignatures(data) {
@@ -114,7 +137,11 @@ Signature.compare = function (a, b) {
 };
 
 function hasTag(body, tag) {
-  return tag.regex.test(body);
+  return tag.regex.test(body || "");
+}
+
+function useGithubApprovalForCr() {
+  return config.useGithubApprovalForCr !== false;
 }
 
 export default Signature;
--- a/README.md
+++ b/README.md
@@ -49,12 +49,17 @@ the pull is considered ready. Conversely, if a pull only touches test code, you
 might put only `qa_req 0` on it to say that it doesn't need to be QAed, since
 it should break tests if there's anything wrong with it.
 
-When you CR a pull, add a comment to it with `CR :emoji:` in it.
+When you CR a pull, you can either approve it on GitHub (using GitHub's native
+**Approve** review action) or add a comment containing `CR :emoji:`.
 (`emoji` is simply a word or words between colons; we often use GitHub emoji,
-which follow this format.) This comment is considered a _signoff_. Pulldasher
-will update the pull's display to indicate that one of the required CRs is
-completed. Similarly, when you QA a pull, add a comment containing
-`QA :emoji:`, and the number of QA signoffs will increase.
+which follow this format.) A GitHub approval and a `CR :emoji:` comment are both
+considered _signoffs_. Pulldasher will update the pull's display to indicate
+that one of the required CRs is completed. Similarly, when you QA a pull, add a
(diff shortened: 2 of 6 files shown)


probablyian:
**QA: logic + webhook wiring verified. Live GitHub E2E not run (DB/infra blockers below).**

**Done**

- `npm test` passes 3/3: APPROVED→CR, dedup against a `CR :emoji:` body, CHANGES_REQUESTED ignored.
- Ran 5 more cases directly against `Signature.parseReview`, all as expected: `COMMENTED`→0 CR; `DISMISSED`→0 CR; `APPROVED` with `body: null`→1 CR and no throw (the `hasTag(body || "")` fix); `APPROVED` + `QA :tada:` body→1 CR (QA also parsed); `APPROVED` with `useGithubApprovalForCr: false`→0 CR.
- Traced both live callers of `parseReview` — both carry `review.state`: the `pull_request_review` webhook (`submitted` → `insertSignatures`, instant) in `controllers/githubHooks.js`, and full refresh via `getReviews`→`pulls.listReviews` in `lib/git-manager.js`. Also confirmed by reading that an approval-derived CR goes inactive after a new commit (stale logic in `git-manager.js`), and DB-stored reviews aren't re-parsed, so there's no stale-`state` path.

**Not done, and why**

The two manual boxes — approve on GitHub fills the CR bubble, dismiss drops it after refresh — weren't verified end-to-end. Test machine has no MySQL/Docker for the app's DB, and real webhook delivery needs a GitHub OAuth app, a public tunnel, and a PR approvable by a non-author (can't approve your own). Not reproducible headlessly.

Net: the parsing logic and the webhook/refresh wiring feeding it are verified; only GitHub's live review-event delivery and the resulting UI bubble are unverified. Recommend a manual approve+dismiss pass on a tracked/staging repo before relying on it, with the **Pull request reviews** webhook event enabled.

kwiens:
Seems reasonable, let's make sure this logs to the db the same as a regular CR.

This is all because github predates GH approvals by many years. We were first!

probablyian:
Confirmed same write path, both triggers:
  - parseReview (approval) and parseComment (regular CR) both build a Signature with the same .data shape — models/signature.js:69 vs :91.
  - Both flow → insertSignatures → insertSignature → new DBSignature → INSERT INTO pull_signatures SET ? (models/db_signature.js:27). Same table, same columns.
  - Webhook path: pull_request_review submitted → insertSignature  (controllers/githubHooks.js:120-125), mirrors the issue_comment path at :91-96.
  - Full-refresh path: git-manager.js:195-199 parses reviews into the same signatures array as comments, then updateAllPullData delete-then-reinsert (db-manager.js:53-57).

probablyian:
@danielbeardsley would you mind giving this a quick check since you're most familiar?

models/signature.js
```diff
@@ -54,9 +54,32 @@ Signature.parseReview = function parseReview(review, repoFullName, pullNumber) {
...
 
-  return parseSignatures(reviewData);
+  const signatures = parseSignatures(reviewData);
+
+  // A GitHub "Approve" review counts as a CR signoff even without a
+  // `CR :emoji:` tag in the body. Skip if the body already produced one.
+  if (
+    useGithubApprovalForCr() &&
+    review.state === "APPROVED" &&
+    !signatures.some((sig) => sig.data.type === "CR")
```
danielbeardsley, line 67: Seems like both of them should count. If I submit a review and someone else places a `CR 👍` comment, I'd expect the pull to have two CRs.

The front-end will de-dupe and only show 1 CR per user I think.

danielbeardsley: Oh! I mis-read this... it's only parsing a single review and thus it's already implemented like I'm suggesting (only adding a signature if the `CR` parsing didn't find one).

Looks good!

danielbeardsley:
CR :+1: 

Nice!

models/signature.js
probablyian: Frontend de-dupe confirmed — `computeSignatures` keeps one sig per user. But the dedupe here is per-review, not cross-review: `parseReview` only sees one review. Your example (my approval + someone else's `CR 👍`) is two parse calls → two CRs by two users → both already count.

This check only suppresses the same user approving *and* writing `CR 👍` in the *same review body* (otherwise a duplicate row the frontend collapses anyway). Can drop it and lean on the frontend de-dupe if you prefer — display is identical.

PR #464, opened 2026-06-17
fix: map cancelled check runs to failure, preserve raw conclusion

mlahargou:
## Summary

- Cancelled GitHub Actions runs (manual cancel or timeout) now map to `failure` instead of `error`, so they block merge like real failures
- The `description` field in `commit_statuses` now stores the **raw GitHub conclusion** (`cancelled`, `failure`, `timed_out`, etc.) instead of the mapped state, so Grafana dashboards can differentiate between true failures and cancelled runs
- Updated both the webhook path (`githubHooks.js`) and the refresh path (`git-manager.js`) for consistency

## Context

GitHub Actions reports skipped/cancelled jobs with conclusion `cancelled`, which was falling into the `default` case in `mapCheckToStatus()` and being recorded as `state='error'`. This polluted CI dashboards with ~800 false error entries per week.

The `description` column is only used for display text in the Pulldasher UI popover — no code branches on its value. Storing the raw conclusion is arguably better UX since it tells users *why* a check failed.

## Test plan

- [ ] Deploy and verify cancelled runs show as `failure` (not `error`) in `commit_statuses`
- [ ] Verify Pulldasher UI shows raw conclusion text in the status popover
- [ ] Verify skipped runs still map to `success`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/controllers/githubHooks.js
+++ b/controllers/githubHooks.js
@@ -155,16 +155,15 @@ const HooksController = {
         dbUpdated = dbManager.updateComment(comment);
       }
     } else if (event === "check_run") {
-      const state = utils.mapCheckToStatus(
-        body.check_run.conclusion || body.check_run.status
-      );
+      const conclusion = body.check_run.conclusion || body.check_run.status;
+      const state = utils.mapCheckToStatus(conclusion);
 
       const status = new Status({
         repo: body.repository.full_name,
         sha: body.check_run.head_sha,
         state: state,
         context: body.check_run.name,
-        description: state,
+        description: conclusion,
         target_url: body.check_run.html_url,
         started_at: body.check_run.started_at,
         completed_at: body.check_run.completed_at,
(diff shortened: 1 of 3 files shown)


mlahargou:
QA 📱 

Before:
image

After:
image

erinemay:
cr 👍 makes sense to me. 
- description now stores the actual gh output instead of a copy of state
- cancelled now returns failure instead of falling through to error

deploy_block 🌵 
Did you mean to add a `timed_out` case mapped to `failure` too? Hard to tell from the description whether you want timeout mapped to failure vs error, or if the dashboards will/can use the new description value of the test result.

mlahargou:
Good call. QA 📱 CR 📱 carry-over

Ref: https://docs.github.com/en/graphql/reference/checks#enum-checkconclusionstate

Issue #467, opened 2026-06-18
Reduce GitHub API consumption to avoid rate-limit outages


probablyian:
## Background

On 2026-06-17, shortly after deploying #463, Pulldasher went down with:

```
GET /orgs/iFixit/members/danielbeardsley - 403
Request quota exhausted for request GET /orgs/{org}/members/{username}
```

### Root cause

The org-membership 403 is a **symptom**, not the cause. That call lives only in `lib/authentication.js` on the login callback — it failed because the *whole* hourly REST quota was already drained, so login couldn't complete (app appears "down").

What drained the quota was the **deploy restart**. On boot, `app.js` runs `refresh.openPulls()`, sending every open pull (across all repos) through `git-manager.parse()`. Each `parse()` is ~8–15 REST requests:

| call | cost |
|---|---|
| `getReviews`, `getIssueComments`, `getPullReviewComments`, `getIssueEvents` | paginated, 1+ each |
| `getIssue`, `getCommit`, `getCombinedStatusForRef` | 1 each |
| `getAllJobRuns` | 1 for workflow runs **+ 1 per workflow run** for jobs |

N open pulls × ~10 calls, back-to-back, blew the 5000/hr quota. **Any restart can trigger this**; #463 was simply the deploy that tipped it over. #463's own code (`models/signature.js` `parseReview`) adds zero API calls.

## Proposed paths (ranked)

1. **ETag conditional requests** — a conditional request returning `304 Not Modified` (`If-None-Match` + `Authorization`) does **not** count against the primary rate limit ([docs](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api)). Octokit does not cache by default. On restart almost every open pull is unchanged, so a per-URL ETag cache turns the startup resync from thousands of billed calls into ~0. Smallest change, biggest win. **Draft PR incoming.**
2. **Trim `getAllJobRuns`** — heaviest per-pull cost (workflow runs + one jobs call per run) and overlaps `getCombinedStatusForRef`. Switch to the Checks API or drop the job-level fetch and lean on the `check_run` webhook.
3. **Don't full-resync on boot** — pulls are already warmed from the DB (`app.js:61-69`); gate or pace the `refresh.openPulls()` cold-start spike.
[...]

PR #468, opened 2026-06-18
Add ETag conditional-request cache to cut GitHub API usage

ianrohde:
Draft for option 1 in #467.

## What

Adds an in-process **ETag conditional-request cache** to the Octokit client. A conditional request returning `304 Not Modified` does **not** count against the primary REST rate limit ([docs](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api)). `lib/etag-cache.js` wraps the Octokit `request` hook to:

- remember each GET response's `ETag`,
- replay it via `If-None-Match` on the next identical request,
- serve the cached body when GitHub responds `304`.

## Why

On boot, `app.js` runs `refresh.openPulls()`, sending every open pull (across all repos) through `git-manager.parse()` — ~8–15 REST calls each. Almost none have changed since the last run, so on a restart this re-fetch is mostly wasted quota and can blow the 5000/hr limit (the incident in #467). With the cache, unchanged resources return `304` and cost nothing.

## How

- `installEtagCache(octokit)` is called once in `git-manager.js`, right after the client is built (composes with the existing throttling plugin).
- GETs only; other methods pass through untouched.
- Bounded, insertion-ordered `Map` used as a rough LRU (default 2000 entries), keyed by the resolved `method + url` so paginated pages cache separately.

## Test plan

- [x] `node --test test/*.test.js` — new `test/etag-cache.test.js` drives a real Octokit with a stub `fetch`, asserting the 200→304 replay path and that the cached `ETag` is sent via `If-None-Match`; non-GETs send no conditional header.
- [x] `npx eslint` on changed files — clean.
- [ ] Manual: deploy, restart, confirm open-pull resync produces `304`s (debug `pulldasher:etag-cache`) and that `core` rate-limit usage drops vs. a cold start.

## Notes / open questions for #467

- Cache is process-local and lost on restart — the first post-restart fetch still spends quota once (then 304s thereafter). A persistent store (Redis) would survive restarts; probably overkill, worth a sentence in #467.
- `maxEntries` default 2000 is a guess; size to open-pull count × endpoints-per-pull.
- Does not change `getAllJobRuns` (#467 item 2) — that's the next lever.

### DIFF
--- a/lib/etag-cache.js
+++ b/lib/etag-cache.js
@@ -1,0 +1,65 @@
+import debug from "./debug.js";
+
+const cacheDebug = debug("pulldasher:etag-cache");
+
+/**
+ * Installs an in-process ETag cache on an Octokit instance.
+ *
+ * GitHub does not count a conditional request against the primary rate limit
+ * when it returns `304 Not Modified`. By remembering the `ETag` of every GET
+ * response and replaying it via `If-None-Match`, unchanged resources cost no
+ * quota. This matters most at startup, where `refresh.openPulls()` re-fetches
+ * every open pull even though almost none have changed since the last run.
+ *
+ * See: https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api
+ *
+ * The cache is a bounded, insertion-ordered Map used as a rough LRU: touching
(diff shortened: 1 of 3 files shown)


probablyian:
I don't think this is necessary

PR #469, opened 2026-06-18
Honor GitHub approvals carried over a clean master merge

ianrohde:
## Summary
Follow-up to #463. A GitHub native **Approve** review (now mapped to a CR signoff) **carries over a clean master merge** — GitHub only dismisses the review when the diff actually changes. Pulldasher dropped the CR anyway, so GitHub showed the PR approved while the Pulldasher CR bubble was empty.

## Root cause
Two paths killed the carried-over approval:

1. **`lib/git-manager.js`** — the full refresh marks any CR/QA whose timestamp predates the head commit inactive. An approval's `submitted_at` legitimately predates a later merge commit, so every refresh re-killed it.
2. **`controllers/githubHooks.js` (`synchronize`)** — a master merge fires only `synchronize`, which blanket-invalidated CR/QA and **never re-derived** review state. With no periodic poll (only startup `refresh.openPulls`), the approval stayed dropped until restart.

## Fix
- Tag the synthesized approval CR with `fromGithubApproval` (`models/signature.js`), in-memory only — not persisted.
- Exempt those signatures from the commit-date staleness check (`lib/git-manager.js`). GitHub is the source of truth: if the review is still `APPROVED` at refresh time, the approval is valid; if a commit should kill it, GitHub returns a non-`APPROVED` state and `parseReview` synthesizes no CR.
- After the `synchronize` invalidate, run a full `refresh.pull` to re-derive review state (`controllers/githubHooks.js`). The immediate invalidate stays for snappy UI; the refresh refills only the still-valid approval.

## Tradeoff
One extra full refresh per push per tracked PR. The ETag cache makes these mostly `304`s, so no rate-limit hit.

## Known edge (unchanged)
A reviewer who *both* approves natively *and* writes `CR :emoji:` → dedup keeps the emoji CR (not exempt), so it still invalidates on merge. Pre-existing dedup behavior, left as-is.

## Test plan
- [x] `npm test` — asserts `fromGithubApproval` true for native approval, false for emoji CR
- [x] `npm run lint`
- [ ] Manual: approve a PR on GitHub, merge master cleanly into the branch, confirm the CR bubble stays filled in Pulldasher
- [ ] Manual: push a real code change after approval, confirm GitHub dismisses and the CR bubble drops

### DIFF
--- a/controllers/githubHooks.js
+++ b/controllers/githubHooks.js
@@ -62,6 +62,9 @@ const HooksController = {
       // Promise that resolves when everything that needs to be done before
       // we call `updatePull` has finished.
       var preUpdate = handleLabelEvents(body);
+      // New commits may carry a GitHub approval over a clean master merge, so
+      // after the cheap DB update we re-derive review state from GitHub.
+      var reconcileAfterUpdate = false;
 
       switch (body.action) {
         case "opened":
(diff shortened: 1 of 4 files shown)


jarstelfox (approved):
CR 🌵

probablyian:
QA 🟢

PR #472, opened 2026-06-23
Make refresh backfill scripts tolerant of transient errors, add per-repo targeting

sctice-ifixit:
## Why

Follow-up to [#457](https://github.com/iFixit/pulldasher/pull/457), which started recording issue author (plus closed reason and assignment date). That PR added the columns and the capture code; the backfill (`refresh-all-issues`) is how those fields get populated for existing issues, so it had to actually finish.

The bulk backfills (`refresh-all-issues` and friends) make thousands of GitHub calls over ~25 minutes, and a single transient `5xx`/network error aborted the whole sweep with no resume. A clean pass failed two consecutive times, and `iFixit/ifixit` kept ending up un-backfilled. This makes the sweep survive transient errors, reports when a repo can't be fetched, and lets us target specific repos.

## What changed

Three layers, smallest blast radius first:

1. **Request layer**: add `@octokit/plugin-retry` so transient `5xx`/network failures auto-retry with backoff (`retry: { retries: 5 }`). The throttle plugin still owns rate limits (403/429), and the two compose.
2. **Fetch fan-out**: `forEachRepo` uses `Promise.allSettled`, so one repo's failure doesn't abort the others. It resolves `{ items, failedRepos }` and logs each skipped repo.
3. **Parse/db queue**: the per-item handlers (`processIssueItem`/`processPullItem`) catch a parse/update failure, log it, and call `next()` to continue instead of crashing the process. This also stops a transient webhook-driven parse error from taking down the live server.

Plus:

- **Completion signal**: at the end of a run the four bulk bins re-report everything that failed (repos that couldn't be fetched, plus individual issues/pulls that couldn't be refreshed) and exit non-zero, so a partial sweep isn't mistaken for a complete one. The per-failure lines are logged inline mid-run too, but they're easy to miss under thousands of debug lines, so they're repeated at the end.
- **Per-repo targeting**: each bulk bin takes optional repo-name args (`node bin/refresh-all-issues iFixit/valkyrie`), for manual QA and for re-running just the repos a sweep skipped. No args means all configured repos. An unconfigured name exits non-zero with a usage line.

Two pre-existing bugs surfaced while testing and got fixed here:

- `lib/db.js` had no `end()`, so every `refresh-*` run threw `db.end is not a function` and exited non-zero even on a clean sweep (which also defeated the new exit-code signal). Added `end()`.
- `controllers/githubHooks.js` called the `debug` factory directly instead of a namespaced logger, so webhook receipt/action/label events never logged. Fixed.

## Tests

A `node --test` suite (no new runtime dep) covers the two isolation paths, the repo selection/targeting, and the retry wiring (a `503` retries then succeeds, a persistent `5xx` gives up after the configured retries, a `404` isn't retried). Writing the retry test caught that `request: { retries }` forces retries on every status (bypassing `doNotRetry`). The count belongs under `retry: { retries }`.

## Validation (QA)

Ran on `pulldasher-dev` against dev metrics:

- Full `iFixit/ifixit` backfill (the run that died on `502`/`500` before): ~6h to completion, 27,046 issues, 0 failures, 0 skipped, 0 aborts. The DB shows ~27k issues with real authors (was `ghost`). 6 rate-limit backoffs were absorbed by the throttle plugin.
- `iFixit/valkyrie` (584 issues), multi-repo runs, and the targeting guard (an unconfigured repo is refused) all came back clean.
- Webhooks: `pulldasher-dev` receives and processes live GitHub `status` events into dev metrics with 0 errors.

### DIFF
--- a/bin/refresh-all-issues
+++ b/bin/refresh-all-issues
@@ -1,11 +1,12 @@
 #!/usr/bin/env node
 import { default as debugInit } from '../lib/debug.js';
 import refresh, { reportFailedRepos } from '../lib/refresh.js';
+import selectReposArg from '../lib/select-repos-arg.js';
 import db from '../lib/db.js';
 
 debugInit('pulldasher:refresh*');
 
-refresh.allIssues()
+refresh.allIssues(selectReposArg())
    .done(function (result) {
       reportFailedRepos(result);
       db.end();
--- a/bin/refresh-all-issues
+++ b/bin/refresh-all-issues
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 import { default as debugInit } from '../lib/debug.js';
-import refresh, { reportFailedRepos } from '../lib/refresh.js';
+import refresh, { reportFailures } from '../lib/refresh.js';
 import selectReposArg from '../lib/select-repos-arg.js';
 import db from '../lib/db.js';
 
@@ -8,6 +8,6 @@ debugInit('pulldasher:refresh*');
 
 refresh.allIssues(selectReposArg())
    .done(function (result) {
-      reportFailedRepos(result);
+      reportFailures(result);
       db.end();
    });
--- a/bin/refresh-all-issues
+++ b/bin/refresh-all-issues
@@ -1,11 +1,12 @@
 #!/usr/bin/env node
 import { default as debugInit } from '../lib/debug.js';
-import refresh from '../lib/refresh.js';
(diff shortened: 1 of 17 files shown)


danielbeardsley (approved):
CR 👍

sctice-ifixit:
QA :tulip: per the description. I also ran all of the `refresh-*` scripts against the `valkyrie` repo without issues.

PR #211, opened 2021-05-05
Change passport-github to passport-github2 and @octokit/rest to v18

addison-grant:
This library seems popular and claims to use a different auth method
that I think will bandaid our current brownout issue.

New package: https://github.com/cfsghost/passport-github
Deprecation notice: (Links to original notice)
https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/

---- 

## Changes made:

1. Removed the usage of the npm package `passport-github` in favor of `passport-github2`
2. Updated the npm package `@octokit/rest` to v18
3. By updating `@octokit/rest` , the function calls in `lib/git-manager.js` had to be updated to reflect the new function calls that call the same endpoints (https://github.com/iFixit/pulldasher/pull/211#issuecomment-833121358). Also had to update the `checkOrgMembershipRequirements` function in `lib/authentication.js`
4. Removed the use of the `rate-limit` function as the plugin `@octokit/plugin-throttling` now allows for the GitHub API object to throttle itself (https://github.com/octokit/plugin-throttling.js#readme)
5. Removed the use of the `paginate` function as `@octokit/rest` v18 now has its own pagination function.

## QA

You can run `update-pulldasher` on cominor with a clone of this repo with this branch checked out, and check out https://pulldasher.cominor.com. Look around pulldasher, let it run throughout the day and make sure the cards reflect the state of their PRs at the end of the day.

Associated issue: #210

PR #473, opened 2026-06-23
Pace bulk refresh against the GitHub API quota

sctice-ifixit:
Connects https://github.com/iFixit/pulldasher/issues/467

## Why

The bulk refreshes (`bin/refresh-*`) share one GitHub token, and one hourly quota, with the live webhook/socket refreshes that keep Pulldasher current. A backfill is a tight loop, so it drains the quota to zero and starves normal operation, taking Pulldasher down. Pulldasher already had `@octokit/plugin-throttling` and `@octokit/plugin-retry`, but both only react at exhaustion (429/5xx), which is too late. This adds a proactive pacer that reserves quota headroom and spreads bulk work across the rate-limit reset window.

## Where this landed

Only the CLI backfill processes pace, and they gate at the point of real API spend (the queue consumer). The server, the webhook/socket path, single-item refreshes, and the startup open-pulls refresh all run an unpaced `noopPacer`. Nothing paces the server, so it can't be starved by its own pacing, and the reserve floor is guaranteed headroom for live traffic rather than something the server has to compete for.

- New `lib/pacer.js`, a per-process pacer. `observe(headers)` records `x-ratelimit-remaining`/`-reset`/`-used` from every response the process makes (so its view stays current). `gate()` waits before a bulk request. The pure `computePace` even-spreads over the reset window (interval = `(reset - now) / (remaining - reserve)`) and hard-pauses below a configurable reserve floor (`config.github.bulkReserve`, default `1000`, 20% of the 5000/hr token). The token-bucket cursor advances by the `x-ratelimit-used` delta between gates, so a pull's multi-call fan-out pushes bulk's next slot out. Pacing tracks real consumption, and bulk yields automatically.
- The pacer is injected per-process, not toggled by a global flag. `refresh.js` grows a `createPacedRefresh()` helper (`createPacer()` + `observeRateLimit(pacer)` + `createRefresh({ pacer })`); the 4 bulk bins call it, the single-item `refresh-pull`/`refresh-issue` bins get the default unpaced instance. `observeRateLimit(pacer)` registers the Octokit `after`/`error` hooks closing over that pacer, with an install-once guard so a second call doesn't stack duplicate hooks.
- Gating happens in the consumer drain and in `pacedPaginate` between fetch pages. `pacedPaginate` only gates when there's a next page (`rel="next"`), so there's no wasted trailing gate that could floor-pause a full reset window after the last page.
- `config.example.js` documents `bulkReserve`.

<details>
<summary>How we got here (two rounds)</summary>

**Round 1: pace the server, at enqueue.** The first version paced inside the server process as well an inside the CLI processes, gating as bulk items were pushed onto the shared `NotifyQueue`. It got there by elimination: gating in the queue's pop handler parks the single shared consumer, so every webhook item queued behind a paused bulk item stalls too (up to the roughly 1h floor-pause). The single queue exists (commit 041e000, 2016) to serialize refreshes and to never refresh the same issue twice concurrently (racing DB writes), so splitting into a separate bulk queue would reintroduce that hazard. Enqueue-side gating kept both guarantees and kept the wait off the consumer. Gating once per item under-paces pulls ~8x (each item spends multiple calls), which is where the `x-ratelimit-used` delta calibration came from. QA'd on `pulldasher-dev`, opened as this PR.

**Round 2: pace the consumer, CLI-only.** Danny's review ([r3478201984](https://github.com/iFixit/pulldasher/pull/473#discussion_r3478201984)) pushed toward pacing the actual API spend rather than the enqueue. Reworked so the pacer gates the consumer drain and only runs in the CLI backfill bins, which made the server's startup `openPulls` unpaced (acceptable: it's startup-only and under a minute). This also dropped the need for enqueue-side gating to protect the consumer, since the server no longer paces at all. The planned `usePacer` module-global setter became the per-call `observeRateLimit(pacer)` + `createPacedRefresh()` shape above (no mutable module global). Commits 67ca46f (consumer-side move), 548eb29 (fold the queue pushes into the drain promise's executor).

</details>

## Validation (QA)

Deployed to `pulldasher-dev`. A paced `refresh-open-pulls iFixit/server-templates` drained 1.5–8s per pull, widening under live webhook bursts (bulk yielding as intended). 58 webhooks received / 53 processed alongside it, 0 restarts, quota moved 4615 → 4561 (the 1000 floor never hit). One transient `401` on the actions-jobs endpoint (unrelated to pacing) was logged and the sweep continued.

### DIFF
--- a/lib/refresh.js
+++ b/lib/refresh.js
@@ -144,8 +144,8 @@ export function makeQueueConsumer(pacer, processItem, deps) {
  */
 function pushOnQueue(queue) {
   return function (githubResponse) {
-    queue.push({ response: githubResponse, onFailure: null });
     return new Promise(function (resolve) {
+      queue.push({ response: githubResponse, onFailure: null });
       queue.push(resolve);
     });
   };
@@ -165,10 +165,10 @@ function pushOnQueue(queue) {
  */
 function pushAllOnQueue(queue, onFailure) {
   return function (githubResponses) {
-    githubResponses.forEach(function (githubResponse) {
-      queue.push({ response: githubResponse, onFailure: onFailure });
-    });
     return new Promise(function (resolve) {
+      githubResponses.forEach(function (githubResponse) {
+        queue.push({ response: githubResponse, onFailure: onFailure });
+      });
       queue.push(resolve);
     });
   };
--- a/lib/refresh.js
+++ b/lib/refresh.js
@@ -4,6 +4,7 @@ import utils from "./utils.js";
 import NotifyQueue from "notify-queue";
 import debug from "./debug.js";
 import Promise from "bluebird";
+import pacer from "./pacer.js";
 
 // Queues for making all refreshes be synchronous, one at a time.
 var issueQueue = new NotifyQueue();
@@ -61,7 +62,8 @@ export default {
  *    and return a promise that is fulfilled when the item is fully processed.
  *
  * Single-item refreshes (webhooks, socket) have no end-of-run report, so no
- * failure collector is attached.
+ * failure collector is attached — and they aren't quota-paced, so live traffic
+ * is never delayed.
  */
 function pushOnQueue(queue) {
   return function (githubResponse) {
@@ -81,13 +83,20 @@ function pushOnQueue(queue) {
  * Each item carries the run's `onFailure` collector (null for single-item
  * webhook/socket refreshes), so a failed parse is recorded for that run's
  * end-of-run report.
+ *
+ * This is the bulk path, so it paces enqueueing against the GitHub quota:
+ * `pacer.gate()` waits for each item's slot *before* pushing it, keeping the
+ * shared queue shallow. The wait happens here, off the queue — so a paced or
+ * floor-paused backfill never blocks the single consumer, and live
+ * webhook/socket refreshes keep draining promptly.
  */
 function pushAllOnQueue(queue, onFailure) {
-  return function (githubResponses) {
+  return async function (githubResponses) {
+    for (const githubResponse of githubResponses) {
+      await pacer.gate();
+      queue.push({ response: githubResponse, onFailure: onFailure });
+    }
     return new Promise(function (resolve) {
-      githubResponses.forEach(function (githubResponse) {
-        queue.push({ response: githubResponse, onFailure: onFailure });
-      });
       queue.push(resolve);
     });
   };
@@ -212,7 +221,9 @@ export function processPullItem(
 // queue or the fixed pop deps, so it stays scoped to the run that enqueued it —
 // a bulk run collects its own failures, while webhook/socket refreshes (which
 // push onFailure null) don't accumulate. Unwrap so the process functions see
-// one response plus their injected collaborators.
+// one response plus their injected collaborators. (Quota pacing happens at
+// enqueue time, in pushAllOnQueue, not here — so a paused backfill can't block
+// the single consumer that webhook refreshes also depend on.)
 issueQueue.pop(function (item, next) {
   if (typeof item === "function") {
     item();
--- a/lib/refresh.js
+++ b/lib/refresh.js
@@ -1,60 +1,137 @@
-import gitManager from "./git-manager.js";
+import gitManager, { observeRateLimit } from "./git-manager.js";
 import dbManager from "./db-manager.js";
 import utils from "./utils.js";
 import NotifyQueue from "notify-queue";
 import debug from "./debug.js";
 import Promise from "bluebird";
-import pacer from "./pacer.js";
-
-// Queues for making all refreshes be synchronous, one at a time.
-var issueQueue = new NotifyQueue();
-var pullQueue = new NotifyQueue();
+import { createPacer, noopPacer } from "./pacer.js";
 
 const refreshDebug = debug("pulldasher:refresh");
 
-export default {
-  ///////   Issues   /////////
+/**
+ * Build a refresh API over its own pair of serial queues. The injected `pacer`
+ * gates the *consumer* — the drain, where the API call actually spends quota
+ * and reports the rate-limit headers — before each bulk work item. It defaults
+ * to a no-op, so the server (webhook/socket refreshes and the startup open-pulls
+ * refresh) runs unpaced. Only the CLI backfill bins pass a real pacer, and in
+ * that process the consumer drains nothing but bulk, so parking it on a paced or
+ * floor-paused slot starves nothing. Single-item webhook/socket refreshes push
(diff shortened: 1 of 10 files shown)


djmetzle (dismissed):
Cool. CR ⏱️

probablyian (dismissed):
CR ⚪

sctice-ifixit:
QA :tulip: per the description.

probablyian (dismissed):

djmetzle (dismissed):
CR 🏃

lib/refresh.js
```diff
@@ -81,13 +83,20 @@ function pushOnQueue(queue) {
  * Each item carries the run's `onFailure` collector (null for single-item
  * webhook/socket refreshes), so a failed parse is recorded for that run's
  * end-of-run report.
+ *
+ * This is the bulk path, so it paces enqueueing against the GitHub quota:
+ * `pacer.gate()` waits for each item's slot *before* pushing it, keeping the
+ * shared queue shallow. The wait happens here, off the queue — so a paced or
+ * floor-paused backfill never blocks the single consumer, and live
+ * webhook/socket refreshes keep draining promptly.
```
danielbeardsley: I don't understand.. 

A backfill is going to happen from the CLI, not the running app. i.e. they are in separate processes. They share the API key of course, but not the queue.

danielbeardsley:
I had a feeling that I remember doing this before.. or something very similar (with a quota buffer).  Hmm... Looks like we deleted it when we upgraded the octokit api: https://github.com/iFixit/pulldasher/pull/211

https://github.com/iFixit/pulldasher/blob/811a7a222b38c43e2a85009f4c282141aba4713e/lib/rate-limit.js

Anyway, the octokit plugin supports throttling across instances but it requires a common storage medium (redis or similar) to coordinate.

lib/refresh.js
danielbeardsley: Is the idea that `pushOnQueue()` is given higher priority than `pushAllOnQueue()`?

sctice-ifixit: > A backfill is going to happen from the CLI, not the running app. i.e. they are in separate processes.

True except for `refresh.openPulls()`, which gets called by the running app on startup. It pushes to the same queue as web hooks and uses `pushAllOnQueue()` to do its work.

> Is the idea that `pushOnQueue()` is given higher priority than `pushAllOnQueue()`?

Via a crude mechanism that sits in front of adding to the queue. `pushAllOnQueue()` waits for the rate limit before adding to the queue, so its waiting happens async independent of the queue processing, which itself happens async of the queue pushes. While it's waiting, it isn't adding things to the queue while web hooks _are_ free to add to the queue during that time. If the bulk producer coroutine isn't rate-limited at all, it will add many items to the queue, the processing of which will be in direct competition with the web hook items.

sctice-ifixit:
> Anyway, the octokit plugin supports throttling across instances but it requires a common storage medium (redis or similar) to coordinate.

I believe this pull's approach should work across instances using the headers we get back from GH as the coordination mechanism.

lib/refresh.js
```diff
@@ -81,13 +83,20 @@ function pushOnQueue(queue) {
...
  */
 function pushAllOnQueue(queue, onFailure) {
-  return function (githubResponses) {
+  return async function (githubResponses) {
+    for (const githubResponse of githubResponses) {
+      await pacer.gate();
+      queue.push({ response: githubResponse, onFailure: onFailure });
```
danielbeardsley: It still feels quite weird to be rate-limiting *pushing* things onto the queue.

i.e. pulling from the queue and executing the api call is what updates the knowledge of the world (the headers) and pushing to the queue is about future commitment.

I haven't hard time imagining if it work work out to have a queue that you're throttling additions to, but draining the queue is actually what gives you the feedback of allowable rates.

Maybe instead we throttle the actual thing that needs throttling to, but have different bulkReserve depending on the source of the action (smaller reserve for things triggered by the web-app and big reserve for cli-triggered actions). The refreshOpenPulls on load doesn't need special treatment as it lasts less than a minute and only happens on startup.

danielbeardsley:
deploy_block :+1: on the discussion above.

lib/refresh.js
sctice-ifixit: It _is_ a little weird. I saw it as a compromise to keep the pacing for the start-up pull refresh without complicating the queue processing. I think it works fine because every response we get from GH still updates the observed rate limit data that the enqueue-gate acts upon, and the processing and enqueueing happen concurrently, so they don't get badly out of sync.

But we can drop pacing the server's pull refresh (like you said, it happens quickly and it isn't _that_ many requests) and mostly isolate pacing to the `refresh-*` scripts that actually run into rate limits. I have the changes made and tested. I'll push them soon.

sctice-ifixit:
un_deploy_block 🌷

lib/refresh.js
danielbeardsley: > enqueueing happen concurrently, so they don't get badly out of sync.

What makes us confident in this?

> I saw it as a compromise to keep the pacing for the start-up pull refresh without complicating the queue processing.

I think pacing the insertion while getting the feedback at the draining seems more complicated and certainly makes it hard to reason about how the system will behave.

probablyian (approved):
No blockers.

Optional follow-up: `lib/git-manager.js:22 console.log(config.github)` dumps the oauth token to stdout. Not introduced here, but adjacent to the diff.

danielbeardsley:
I read deeply through this and I approve!

However, unless I'm missing something, we'd still want to use a larger `bulkReserve` for cli-launched scripts vs the config used for the server.

Right now, I *think* it's possible that a cli script can starve out the front-end (Two separate pacers). If they have different bulk-reserves then the the cli refresh will only proceed while the frontend is idle and the token bucket builds back up.

sctice-ifixit:
> Right now, I _think_ it's possible that a cli script can starve out the front-end (Two separate pacers).

Right now, the CLI is the only path that is paced, and it will never (modulo bugs) permit using any of the last 1000 quota available per hour. So once a CLI script sees there are less than 1000 requests remaining, it will pause until the quota is refilled (GH tells us when that will be in the headers). So the front-end should always have at least 1000 requests to work with in any given hour (in practice more), and it always goes as fast as it can. If we think 1000 isn't conservative enough, we can up it.

danielbeardsley:
> Right now, the CLI is the only path that is paced

Ah! Good point:

> function createRefresh({ pacer = noopPacer }

The default is a noopPacer and thus pacing is opt-in and only the cli commands opt-in.

CR :+1:

sctice-ifixit:
QA :tulip: I updated the description to match where we ended up and re-ran QA per the description.

